PR #4, opened 2021-03-29
Development refresh command(s).

markdorison:
## Description
* Added robo commands including `composer robo dev:refresh`:
    > Completely refreshes a development environment including running 'composer install', starting Lando, downloading a database dump, importing it, running 'drush deploy', disabling front-end caches, and providing a login link.
* Moved database:download functionality to this repo. Replaced an AWS CLI implementation with an AWS PHP library.

## Motivation / Context
Provide a better getting started experience for developers.

Test with PR's that reference this PR (see PR activity below)

## Screenshots
Included in associated PRs.

## Documentation
Yes.

### DIFF
--- a/.sites.config.yml
+++ b/.sites.config.yml
@@ -1,10 +1,10 @@
 ---
 default:
   theme_build:
-  - theme_path: src/Robo
-    theme_build_commands:
-      - pwd
-      - echo "This is where we pretend to build a theme!"
-  - theme_path: .github/
-    theme_build_commands:
-      - pwd
+    - theme_path: src/Robo
+      theme_build_commands:
+        - pwd
+        - echo "This is where we pretend to build a theme!"
+    - theme_path: .github/
+      theme_build_commands:
+        - pwd
--- a/.sites.config.yml
+++ b/.sites.config.yml
@@ -1,0 +1,10 @@
+---
+default:
+  theme_build:
+  - theme_path: src/Robo
+    theme_build_commands:
+      - pwd
+      - echo "This is where we pretend to build a theme!"
+  - theme_path: .github/
+    theme_build_commands:
+      - pwd
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -113,13 +113,17 @@ class DevelopmentModeCommands extends Tasks
         $latestDatabaseDump = array_pop($objects);
         $dbFilename = $latestDatabaseDump->getKey();
 
-        $result = $s3->GetObject([
-            'Bucket' => $bucket,
-            'Key' => $dbFilename,
-        ]);
-        $fp = fopen($dbFilename, 'wb');
-        stream_copy_to_stream($result->getBody()->getContentAsResource(), $fp);
-        $this->say('Database dump file downloaded >>> ' . $dbFilename);
+        if (file_exists($dbFilename)) {
+            $this->say("Skipping download. Latest database dump file exists >>> $dbFilename");
+        } else {
+            $result = $s3->GetObject([
+                'Bucket' => $bucket,
+                'Key' => $dbFilename,
+            ]);
+            $fp = fopen($dbFilename, 'wb');
+            stream_copy_to_stream($result->getBody()->getContentAsResource(), $fp);
+            $this->say("Database dump file downloaded >>> $dbFilename");
+        }
         return $dbFilename;
     }
 
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -5,11 +5,11 @@ namespace ChqRobo\Robo\Plugin\Commands;
 use ChqRobo\Robo\Plugin\Traits\SitesConfigTrait;
 use AsyncAws\S3\S3Client;
 use DrupalFinder\DrupalFinder;
-use Drupal\Component\Serialization\Yaml;
 use Robo\Exception\TaskException;
 use Robo\Result;
 use Robo\Robo;
 use Robo\Tasks;
+use Symfony\Component\Yaml\Yaml;
 
 /**
  * Robo commands related to changing development modes.
@@ -216,7 +216,7 @@ class DevelopmentModeCommands extends Tasks
      */
     protected function landoUri($siteDir): string
     {
-        $landoConfig = Yaml::decode(file_get_contents('.lando.yml'));
+        $landoConfig = Yaml::parseFile('.lando.yml');
         $uri = $landoConfig['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI'] ?? null;
         if ($uri) {
             return $uri;
@@ -330,12 +330,12 @@ class DevelopmentModeCommands extends Tasks
             ->run();
 
         $this->say("enablig twig.debug in development.services.yml.");
-        $devServices = Yaml::decode(file_get_contents($this->devServicesPath));
+        $devServices = Yaml::parseFile($this->devServicesPath);
         $devServices['parameters']['twig.config'] = [
             'debug' => true,
             'auto_reload' => true,
         ];
-        file_put_contents($this->devServicesPath, Yaml::encode($devServices));
+        file_put_contents($this->devServicesPath, Yaml::dump($devServices));
 
         $this->say("disabling render and dynamic_page_cache in settings.local.php.");
         $result = $this->collectionBuilder()
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -66,13 +66,13 @@ class DevelopmentModeCommands extends Tasks
         $this->io()->title('development environment refresh. 🦄✨');
         $result = $this->taskComposerInstall()->run();
         $result = $this->taskExec('lando')->arg('start')->run();
-        $result = $this->databaseRefreshLando($siteName);
         // There isn't a great way to call a command in one class from another.
         // https://github.com/consolidation/Robo/issues/743
         // For now, it seems like calling robo from within robo works.
         $result = $this->taskExec("composer robo theme:build $siteName")
             ->run();
         $result = $this->frontendDevEnable($siteName, ['yes' => true]);
+        $result = $this->databaseRefreshLando($siteName);
         $result = $this->drupalLoginLink($siteName);
         return $result;
     }
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -64,10 +64,15 @@ class DevelopmentModeCommands extends Tasks
      */
     public function devRefresh($siteName = 'default'): Result
     {
-        $this->io()->title('developer magic. 🦄');
-        $this->taskComposerInstall()->run();
-        $this->taskExec('lando')->arg('start')->run();
+        $this->io()->title('development environment refresh. 🦄✨');
+        $result = $this->taskComposerInstall()->run();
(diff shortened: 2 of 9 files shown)


adamzimmermann:
I see a request for review, but it is still a draft and has no testing steps. Is this ready or should I hold off still?

My only comment after only reading the PR description, is should this turn on Lando too? Turning that on seems separate from getting the site setup, but I guess it can't hurt even if it is already started?

markdorison:
@adamzimmermann I was looking for any high-level feedback at this stage, like the point you mentioned except that it [already does make sure Lando has been started](https://github.com/ChromaticHQ/chq-robo/pull/4/files#diff-f97502affc46a87c9b2819746c440a2b3bd57e1738085378386a5da44066cf28R68)!

adamzimmermann (commented):
Nothing jumps out, I think it might be best to demo this on our devops call tomorrow. I really like what I see so far.


I have some nitpick questions, but I'll hold off on those for the moment.

.sites.config.yml
```diff
@@ -0,0 +1,10 @@
+---
+default:
+  theme_build:
+    - theme_path: src/Robo
```
adamzimmermann, line 4: Can we provide a better example?

adamzimmermann: Actually, how is this file even used?

src/Robo/Plugin/Traits/SitesConfigTrait.php
```diff
@@ -0,0 +1,98 @@
...
+
+namespace ChqRobo\Robo\Plugin\Traits;
+
+use Robo\Exception\TaskException;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * Trait to provide site configuration functionality to Robo commands.
+ */
+trait SitesConfigTrait
```
adamzimmermann, line 11: 🎩

.sites.config.yml
adamzimmermann: I feel like I'm missing something, as I see better examples in the file above.

adamzimmermann (approved):
Looking good. Excited to get this out there!

.sites.config.yml
markdorison: This file is present so we can [run tests in this repo](https://github.com/ChromaticHQ/chq-robo/blob/c91e5a9021499f8b284444add97303880914ff4c/.github/workflows/ci.yml#L50-L51) and verify the code is looking in the right place. The example file is what should be referenced for setting this up on a real site.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
+<?php
```
markdorison, line 1: @adamzimmermann Did you get a chance to review this file? I just wanted to flag it because GitHub keeps collapsing it by default.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+            }
+        }
+
+        if (!$bucket = $this->getConfig('database_s3_bucket', $siteName)) {
+            throw new TaskException($this, "database_s3_bucket value not set for '$siteName'.");
+        }
+
+        $s3 = new S3Client();
+        $objects = $s3->listObjectsV2(['Bucket' => $bucket]);
+        $objects = iterator_to_array($objects);
```
adamzimmermann, line 110: I didn't know about this. Very cool.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+        // Ensure objects are sorted by last modified date.
+        usort($objects, fn($a, $b) => $a->getLastModified()->getTimestamp() <=> $b->getLastModified()->getTimestamp());
```
adamzimmermann, line 112: 😎

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+        $latestDatabaseDump = array_pop($objects);
+        $dbFilename = $latestDatabaseDump->getKey();
+
+        if (file_exists($dbFilename)) {
+            $this->say("Skipping download. Latest database dump file exists >>> $dbFilename");
+        } else {
```
adamzimmermann, line 118:
```suggestion
        }
        else {
```
Should we enable code style checks on this repo?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+     *   The Drupal site directory name.
+     * @param bool $lando
+     *   Use lando to call drush, else call drush directly.
+     *
+     * @aliases uli
+     *
+     * @return \Robo\Result
+     *   The result of the set of tasks.
+     */
+    public function drupalLoginLink($siteDir = 'default', $lando = true): Result
```
adamzimmermann, line 205:
```suggestion
    public function drupalLoginLink($siteDir = 'default', $lando = true): Result {
```
We want this, right?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+    {
+        $this->io()->section("create login link.");
+        if ($lando) {
+            $uri = $this->landoUri($siteDir);
```
adamzimmermann, line 209: I see we are using 4 spaces instead of 2. Is that intentional?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: I see you are doing this elsewhere too. Is this intentional?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: I see this elsewhere too. Is that intentional?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+                throw new TaskException(
+                    $this,
+                    'Unable to determine URI. Multi-site detected, but you did not specify a site name.',
+                );
+            }
+            // Detect multi-site configurations.
+            $siteDomains = array_filter($landoCfg['proxy']['appserver'], fn($domain) =>
+                strpos($domain, $siteDir) !== false);
+            if (count($siteDomains) > 1) {
+                throw new TaskException($this, 'More than one possible URI found.');
```
adamzimmermann: Should we display the matches? Or is this an extreme edge case?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+            } elseif (count($siteDomains) == 1) {
+                $domain = array_pop($siteDomains);
+                return "http://$domain";
+            }
+        } elseif ($uri = $landoCfg['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI'] ?? null) {
```
adamzimmermann, line 254: Is it safe to assume `$landoCfg['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI']` will always exist? Should we add some checks around this?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -0,0 +1,416 @@
...
+     *   The result of the set of tasks.
+     *
+     * @see https://www.drush.org/deploycommand
+     */
+    protected function drushDeployLando($siteDir = 'default'): Result
+    {
+        $this->io()->section('drush deploy.');
+        return $this->taskExecStack()
+            ->dir("web/sites/$siteDir")
+            ->exec("lando drush deploy --yes")
```
adamzimmermann, line 311: Did you resolve the double deploy issue we saw this morning? I assume yes?

adamzimmermann (approved):
Thanks for calling that out. I think I missed it. Nothing that worries me really, just a lot of questions and gold plating.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: [Code style checks are enabled](https://github.com/ChromaticHQ/chq-robo/blob/c91e5a9021499f8b284444add97303880914ff4c/robo.yml#L12-L13). We are following [PSR-12](https://www.php-fig.org/psr/psr-12/). I don't really have a strong feeling about this but it seemed weird to follow Drupal coding standards in a non-Drupal repo.

For example, [Drush](https://github.com/drush-ops/drush/blob/10.x/CONTRIBUTING.md#coding-style) follows [PSR-2](https://www.php-fig.org/psr/psr-12/) which has been deprecated in favor of PSR-12.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: See comment above.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: See PSR-12 comment above.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: We are checking; with the [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php)!

image

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: It wasn't my code, but it was an extra config import.

It's being caused by our Lando configuration. We should discuss the best way to resolve.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: Ah, I just didn't side scroll far enough to see that. 🤦🏼

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: Interesting. I guess I'm willing to give it a try, but here are my two arguments against it FWIW:

- We are pretty much using this exclusively with Drupal, so it's very much Drupal related.
- It really throws me off when reviewing code. How was it to write it in this format?

markdorison: > We are pretty much using this exclusively with Drupal, so it's very much Drupal related.

I mean, can it get more "Drupal-related" than Drush? They aren't using Drupal coding standards.

> It really throws me off when reviewing code. How was it to write it in this format?

It was fine. When I fell into incorrect habits, [PHPCS would clean it up](https://github.com/ChromaticHQ/chq-robo/blob/c91e5a9021499f8b284444add97303880914ff4c/src/Robo/Plugin/Commands/CICommands.php#L102-L113) for me!

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: That sounds like a nice improvement to make.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: > I mean, can it get more "Drupal-related" than Drush? They aren't using Drupal coding standards.

I mean ya haha. I guess I would question this choice too, but 🤷🏼 ?

👍🏼

PR #7, opened 2021-04-02
Support sites with no theme build steps.

markdorison:
## Description
Support sites with no theme build steps.

## Motivation / Context
Some sites commit a fully-built theme.

Tested locally. See @markdorison for details.

## Screenshots
N/A

## Documentation
N/A

### DIFF
--- a/src/Robo/Plugin/Commands/ThemeCommands.php
+++ b/src/Robo/Plugin/Commands/ThemeCommands.php
@@ -35,8 +35,13 @@ class ThemeCommands extends Tasks
      */
     public function themeBuild($siteName = 'default'): Result
     {
-        $themeBuildConfiguration = $this->getConfig('theme_build', $siteName);
-        $this->io()->title("theme build");
+        $result = $this->io()->title("theme build");
+        try {
+            $themeBuildConfiguration = $this->getConfig('theme_build', $siteName);
+        } catch (TaskException $e) {
+            $this->say("'$siteName' theme_build confguration not set.");
+            return $this->taskExec('echo skipping')->run();
+        }
         foreach ($themeBuildConfiguration as $themeConfig) {
             $themePath = $themeConfig['theme_path'];
             $this->io()->section("building theme at $themePath");


src/Robo/Plugin/Commands/ThemeCommands.php
```diff
@@ -35,8 +35,13 @@ public function __construct()
...
     public function themeBuild($siteName = 'default'): Result
     {
-        $themeBuildConfiguration = $this->getConfig('theme_build', $siteName);
-        $this->io()->title("theme build");
+        $result = $this->io()->title("theme build");
+        try {
+            $themeBuildConfiguration = $this->getConfig('theme_build', $siteName);
+        } catch (TaskException $e) {
+            $this->say("'$siteName' theme_build confguration not set.");
+            return $this->taskExec('echo skipping')->run();
```
adamzimmermann, line 43: Will this just print `skipping` then? Is there a more robo way to do this?

src/Robo/Plugin/Commands/ThemeCommands.php
```diff
@@ -35,8 +35,13 @@ public function __construct()
      */
     public function themeBuild($siteName = 'default'): Result
     {
-        $themeBuildConfiguration = $this->getConfig('theme_build', $siteName);
-        $this->io()->title("theme build");
+        $result = $this->io()->title("theme build");
+        try {
+            $themeBuildConfiguration = $this->getConfig('theme_build', $siteName);
+        } catch (TaskException $e) {
```
adamzimmermann, line 41: Will this always be the exception class or do we need to catch anything more broad? e.g. `\Exception`? This is more for my edification than a serious concern.

adamzimmermann (approved):

src/Robo/Plugin/Commands/ThemeCommands.php
markdorison: It's our exception so we can make it whatever we want.

https://github.com/ChromaticHQ/chq-robo/blob/ad33f7731f25dccfd916b08e23cf9f151c0af2fd/src/Robo/Plugin/Traits/SitesConfigTrait.php#L85

src/Robo/Plugin/Commands/ThemeCommands.php
markdorison: The problem is I need to return a `Result` object from this method (well, I would _like_ to) and if I bail because I don't get a theme configuration back, then we weren't actually doing anything robo-wise that returns a `Result` object.

adamzimmermann: Could say this the robo way, then just create a `Result(TRUE)` object, or however you would do that?

markdorison: It's a little more complicated than that: https://github.com/consolidation/Robo/blob/063a9db302b01d1a7ffa52a8823fceb48312fc9b/src/Result.php#L21-L36

adamzimmermann: Ya, I figured my idea might be too good to be true, but was hoping it sparked an idea for you. 🤷🏼

PR #8, opened 2021-04-02
Widen robo/drush version support.

markdorison:
## Description
Widen drush version support.

## Motivation / Context
Allow this version chq-robo to be installed on older sites where robo 1.x is needed for Drush. This will also allow the latest version of Drush ^8.4 to be installed. When we installed robo 2.x on D7 sites it caused Drush to downgrade to 8.0.4 (I think) because later versions [require a symfony version](https://github.com/drush-ops/drush/commit/cc75da0ff593777b994abfad6ca8ef0407e61374) that conflicts with robo 2.x.

The biggest blocker to this has since been replaced in the code so there should not be a technical limitation in this project's code 🤞. It would be nice to not have to support robo 1.x, but the functionality on D7 sites is too valuable.

Not _all_ chq-robo functionality will work as expected.

Tested with private D7 project. (If you have access you will see the reference in the activity log below).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,13 +10,13 @@
         }
     ],
     "require": {
-        "consolidation/robo": "^2",
+        "consolidation/robo": "^1.4 || ^2",
         "squizlabs/php_codesniffer": "^3.5",
         "phpstan/phpstan": "^0.12.77",
         "phpstan/phpstan-deprecation-rules": "^0.12.6",
         "async-aws/s3": "^1.8",
         "php": ">=7.4",
-        "drush/drush": ">=8.0",
+        "drush/drush": "^8 || ^10",
         "webflo/drupal-finder": "^1.2"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -14,9 +14,10 @@
         "squizlabs/php_codesniffer": "^3.5",
         "phpstan/phpstan": "^0.12.77",
         "phpstan/phpstan-deprecation-rules": "^0.12.6",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -10,13 +10,14 @@
...
+        "consolidation/robo": "^1.4 || ^2",
         "squizlabs/php_codesniffer": "^3.5",
         "phpstan/phpstan": "^0.12.77",
         "phpstan/phpstan-deprecation-rules": "^0.12.6",
-        "drush/drush": "^10.4",
         "async-aws/s3": "^1.8",
-        "php": ">=7.4"
+        "php": ">=7.4",
+        "drush/drush": "^8 || ^10",
+        "webflo/drupal-finder": "^1.2"
```
adamzimmermann, line 20: I assume newer versions of drupal or symphony add this for us?

adamzimmermann (approved):
Should this go into the `1.x` branch, or is `2.x` intentional? I haven't been following your branch patterns too closely.

markdorison:
> Should this go into the 1.x branch, or is 2.x intentional? I haven't been following your branch patterns too closely.

If we had to do write different versions of our code to support the same functionality, then that could be a good approach, but here we are able to support all of these versions on one branch so I don't see any reason to not include that on 2.x which contains new functionality that 1.x does not have.

composer.json
markdorison: No. This was just missing from our dependencies. It was missed because it is often already included on projects.

PR #10, opened 2021-04-07
Allow overriding of S3 region.

markdorison:
## Description
Allow overriding of S3 region.

## Motivation / Context
Database download would fail if bucket was not in default region.

Tested locally.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -101,7 +101,9 @@ class DevelopmentModeCommands extends Tasks
             }
         }
 
-        $s3 = new S3Client();
+        $s3 = new S3Client([
+            'region' => $this->s3RegionForSite($siteName),
+        ]);
         $objects = $s3->listObjectsV2($this->s3BucketRequestConfig($siteName));
         $objects = iterator_to_array($objects);
         // Ensure objects are sorted by last modified date.
@@ -113,7 +115,7 @@ class DevelopmentModeCommands extends Tasks
             $this->say("Skipping download. Latest database dump file exists >>> $dbFilename");
         } else {
             $result = $s3->GetObject([
-                'Bucket' => $bucket,
+                'Bucket' => $this->s3BucketForSite($siteName),
                 'Key' => $dbFilename,
             ]);
             $fp = fopen($dbFilename, 'wb');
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -163,26 +165,63 @@ protected function configureAwsCredentials(string $awsConfigDirPath, string $aws
...
+    /**
+     * Get S3 region for site.
+     *
+     * @param string $siteName
+     *   The site name.
+     *
+     * @return string
+     *   An S3 region.
+     */
+    protected function s3RegionForSite(string $siteName): string
```
adamzimmermann, line 213: Should this leverage `s3BucketRequestConfig`?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -163,26 +165,63 @@ protected function configureAwsCredentials(string $awsConfigDirPath, string $aws
...
         try {
-            $s3KeyPrefix = $this->getConfig('database_s3_key_prefix_string', $siteName);
-            $this->say("'$siteName' S3 Key prefix: '$s3KeyPrefix'");
-            $s3ConfigArray['Prefix'] = $s3KeyPrefix;
+            $region = $this->getConfig('database_s3_region', $siteName);
+            $this->say("'$siteName' database_s3_region set to $region.");
         } catch (TaskException $e) {
-            $this->say("No S3 Key prefix found for $siteName.");
+            // Set default region if one is not set.
+            $defaultRegion = 'us-east-1';
```
adamzimmermann: Should this be a constant?

adamzimmermann (approved):

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: I don't follow.

adamzimmermann: I was thinking that `s3BucketRequestConfig` was getting all config in a generic way, but I think that assumption was wrong. It's getting config but then building something unique.

mmatsoo (approved):
👍

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: s3BucketRequestConfig is also config needed for a request, whereas this is needed to instantiate the client.

PR #11, opened 2021-04-07
Added catch to handle no db configured for download.

markdorison:
## Description
Added catch to handle no db configured for download.

## Motivation / Context
If a site does not have a db bucket configured, it should not fail the entire Tugboat refresh.

Tested locally.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -336,7 +336,13 @@ class DevelopmentModeCommands extends Tasks
     {
         $this->io()->title('refresh tugboat databases.');
         foreach ($this->getAllSitesConfig() as $siteName => $siteInfo) {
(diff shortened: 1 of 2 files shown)


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -336,7 +336,13 @@ public function databaseRefreshTugboat(): Result
     {
         $this->io()->title('refresh tugboat databases.');
         foreach ($this->getAllSitesConfig() as $siteName => $siteInfo) {
-            $dbPath = $this->databaseDownload($siteName);
+            try {
+                $dbPath = $this->databaseDownload($siteName);
+            } catch (TaskException $e) {
+                $this->yell("$siteName: No database configured. Download/import skipped.");
+                // @todo: Should we run a site-install by default?
```
adamzimmermann, line 343: A topic for the devops call next week perhaps?

adamzimmermann (approved):

PR #17, opened 2021-05-07
Adds support for disabling the AdvAgg module.

chrisfree:
## Description
I discovered that on CHQ running of the `composer robo frontend:dev-enable` command did not disable the advanced aggregation module, which means the intended outcome for front-end devs is never reached.

## Motivation / Context
Some sites use the [AdvAgg](https://www.drupal.org/project/advagg) module and thus will need to have it also disabled whenever the `composer robo frontend:dev-enable` command is run.

Pointed a local Drupal site to this branch and tested `composer robo frontend:dev-enable` and `composer robo:dev-refrsh`. Now the follow snippet is added to the bottom of the resulting `settings.local.php` file:

```php
/**
 *  Disable advagg module.
 */
$config['advagg.settings']['enabled'] = FALSE;
```

## Questions
* Is there a better way to handle this? 
* Will sites that don't use AdvAgg simply ignore this config override? I think the answer is yes.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -462,7 +462,7 @@ class DevelopmentModeCommands extends Tasks
             ->append(true)
             ->line('')
             ->line('/**')
-            ->line(' *  Disable advagg module.')
+            ->line(' *  If advagg module is present, disable its functionality.')
             ->line(' */')
             ->line('$config[\'advagg.settings\'][\'enabled\'] = FALSE;')
             ->run();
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -458,6 +458,13 @@ public function frontendDevEnable($siteDir = 'default', array $opts = ['yes|y' =
             ->taskReplaceInFile($devSettingsPath)
             ->from('# $settings[\'cache\'][\'bins\'][\'page\'] = ')
             ->to('$settings[\'cache\'][\'bins\'][\'page\'] = ')
+            ->taskWriteToFile($devSettingsPath)
+            ->append(true)
+            ->line('')
+            ->line('/**')
+            ->line(' *  Disable advagg module.')
```
markdorison:
```suggestion
            ->line(' *  If advagg module is present, disable its functionality.')
```

markdorison (changes requested):

chrisfree:
@markdorison ready for final review

markdorison (approved):

PR #24, opened 2021-06-10
Fix multi-site URI check.

markdorison:
## Description
Fix multi-site URI check.

## Motivation / Context
Failing when multiple of the URIs contain, but do not start with the $siteDir.

Tested locally.

## Documentation
N/A

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -314,8 +314,10 @@ class DevelopmentModeCommands extends Tasks
                 );
             }
             // Detect multi-site configurations.
+            // We look for the $siteDir to be at the beginning of the appserver
+            // proxy URL.
             $siteDomains = array_filter($landoCfg['proxy']['appserver'], fn($domain) =>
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -314,8 +314,10 @@ protected function landoUri($siteDir): string
                 );
             }
             // Detect multi-site configurations.
+            // We look for the $siteDir to be at the beginning of the appserver
+            // proxy URL.
             $siteDomains = array_filter($landoCfg['proxy']['appserver'], fn($domain) =>
-                strpos($domain, $siteDir) !== false);
+                strpos($domain, $siteDir) == 0);
```
adamzimmermann:
```suggestion
                strpos($domain, $siteDir) === 0);
```
We still want strict comparison right?

adamzimmermann (approved):

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: You beat me to it!

PR #26, opened 2021-07-29
Add Twig linting.

adamzimmermann:
## Description
- Adds Twig code style checks to the code style step.
- Adds automated Twig code style fixing to automated fixing step.

## Motivation / Context
- We want to ensure standardized Twig template formatting.
- https://github.com/ChromaticHQ/chromatichq.com/issues/3314

Run `composer cs-check` or `composer cs-fix` in a repo that is using this branch.

## Screenshots
127340685-86ffa043-1e74-40d7-a035-17a148ad0ddd

## Documentation
Added example to template file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,6 +11,3 @@ indent_size = 4
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
-[composer.{json,lock}]
-indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# @see http://editorconfig.org/
+
+# This is the top-most .editorconfig file; do not search in parent directories.
+root = true
+
+# All files.
+[*]
+end_of_line = LF
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[composer.{json,lock}]
+indent_size = 4
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor
 composer.lock
+.idea
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -99,13 +99,15 @@ class CICommands extends Tasks
         $standards = implode(',', $this->getCodingStandards());
         $extensions = implode(',', $this->phpcsCheckExtensions);
         $ignorePaths = implode(',', $this->phpcsIgnorePaths);
-        return $this->taskExecStack()
-            ->stopOnFail()
-            ->exec('vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer')
-            ->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
-                --ignore=$ignorePaths $this->customCodePaths")
-            ->exec("vendor/bin/twig-cs-fixer lint $this->customCodePaths")
-            ->run();
+        /** @var \Robo\Task\CommandStack $stack */
+        $stack = $this->taskExecStack()->stopOnFail();
+        $stack->exec('vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer');
+        $stack->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
+                --ignore=$ignorePaths $this->customCodePaths");
+        if ($this->lintTwigFiles) {
+            $stack->exec("vendor/bin/twig-cs-fixer lint $this->customCodePaths");
+        }
+        return $stack->run();
     }
 
     /**
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -35,6 +35,13 @@ class CICommands extends Tasks
     protected $customCodePaths;
 
     /**
+     * Boolean indicating if Twig files should be linted.
+     *
+     * @var bool
+     */
+    protected $lintTwigFiles;
+
+    /**
      * RoboFile constructor.
      */
     public function __construct()
@@ -45,6 +52,7 @@ class CICommands extends Tasks
         $this->phpcsCheckExtensions = Robo::config()->get('phpcs_check_extensions');
         $this->phpcsIgnorePaths = Robo::config()->get('phpcs_ignore_paths');
         $this->customCodePaths = implode(' ', $this->getCustomCodePaths());
+        $this->lintTwigFiles = Robo::config()->get('twig_lint_enable', true);
     }
 
     /**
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 vendor
 composer.lock
+.idea
```
adamzimmermann, line 3: Added this for PHPStorm.

adamzimmermann:
- [x] I still need to add the ability to toggle this functionality. Standby.

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -34,6 +34,13 @@ class CICommands extends Tasks
      */
     protected $customCodePaths;
 
+    /**
+     * Boolean indicating if Twig files should be linted.
```
agarzola:
```suggestion
     * Boolean indicating whether Twig files should be linted.
```

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+[*]
+end_of_line = LF
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[composer.{json,lock}]
+indent_size = 4
```
markdorison: Is this an override? Do we need this since it matches the indent_size above?

agarzola (approved):
I haven’t tested this directly, but the code changes look reasonable to me and the accompanying PR on CHQ validates that it works as expected. LGTM

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -91,12 +99,15 @@ public function jobCheckCodingStandards(): Result
...
-            ->stopOnFail()
-            ->exec('vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer')
-            ->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
-                --ignore=$ignorePaths $this->customCodePaths")
-            ->run();
+        /** @var \Robo\Task\CommandStack $stack */
+        $stack = $this->taskExecStack()->stopOnFail();
+        $stack->exec('vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer');
+        $stack->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
+                --ignore=$ignorePaths $this->customCodePaths");
```
markdorison:
```suggestion
            --ignore=$ignorePaths $this->customCodePaths");
```

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -112,12 +123,16 @@ public function jobFixCodingStandards(): Result
...
-            ->stopOnFail()
-            ->exec('vendor/bin/phpcbf --config-set installed_paths vendor/drupal/coder/coder_sniffer')
-            ->exec("vendor/bin/phpcbf --standard=$standards --extensions=$extensions \
-                --ignore=$ignorePaths $this->customCodePaths")
-            ->run();
+        /** @var \Robo\Task\CommandStack $stack */
+        $stack = $this->taskExecStack()->stopOnFail();
+        $stack->exec('vendor/bin/phpcbf --config-set installed_paths vendor/drupal/coder/coder_sniffer');
+        $stack->exec("vendor/bin/phpcbf --standard=$standards --extensions=$extensions \
+                    --ignore=$ignorePaths $this->customCodePaths");
```
markdorison:
```suggestion
                --ignore=$ignorePaths $this->customCodePaths");
```

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -91,12 +99,15 @@ public function jobCheckCodingStandards(): Result
...
-            ->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
-                --ignore=$ignorePaths $this->customCodePaths")
-            ->run();
+        /** @var \Robo\Task\CommandStack $stack */
+        $stack = $this->taskExecStack()->stopOnFail();
+        $stack->exec('vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer');
+        $stack->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
+                --ignore=$ignorePaths $this->customCodePaths");
+        if ($this->lintTwigFiles) {
+            $stack->exec("vendor/bin/twig-cs-fixer lint $this->customCodePaths");
```
markdorison, line 108: Since we are adding this, we may want to add an additional alias to the job instead of just `phpcs`. I can't add a suggestion/comment on that line (92), but it would look like this:

```
* @aliases phpcs,twigcs
```

markdorison, line 108: I am also thinking about it if it would be better to break into two separate jobs but that is not a blocker for this. Need to think on it further.

markdorison (commented):
Minor suggestions. This is looking good!

.editorconfig
adamzimmermann: Ah good call. Copy paste got me. (I stole this from Drupal).

src/Robo/Plugin/Commands/CICommands.php
adamzimmermann: I like that we are thinking about this.

I guess I never even thought about this, as I only interact with the end command of `composer cs-check`, which makes no mention of PHP or Twig.

Does that change your opinion?

I went ahead and added the alias for now though.

markdorison: `composer cs-check` is a repo shortcut to call this job, so it is 1:1. I am trying to consider scenarios where someone working locally might want to lint twig without linting PHP either for speed reasons or because they know PHP is in a failed state. The way this is currently implemented if an earlier step (like PHP linting) fails it will skip the rest.

markdorison: I also just realized in my testing that implemented in this way, twig linting will run on this repo, which doesn't use twig. Again, not a blocker, but that might tip me over to splitting it up. 🤔

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -45,6 +52,7 @@ public function __construct()
         $this->phpcsCheckExtensions = Robo::config()->get('phpcs_check_extensions');
         $this->phpcsIgnorePaths = Robo::config()->get('phpcs_ignore_paths');
         $this->customCodePaths = implode(' ', $this->getCustomCodePaths());
+        $this->lintTwigFiles = Robo::config()->get('twig_lint_enable', true);
```
markdorison, line 55: @adamzimmermann If we are going to default to `true`, I think we should set `twig_lint_enable` to `false` for this repo in `robo.yml`.

src/Robo/Plugin/Commands/CICommands.php
adamzimmermann: Ya, I can see how it would be good to separate it. Let me sort out the other issues then we can decide if we do that in this PR or in a follow-up one.

src/Robo/Plugin/Commands/CICommands.php
adamzimmermann: ✅

src/Robo/Plugin/Commands/CICommands.php
adamzimmermann: The blocker in the CHQ PR has been resolved and I have added the ability to toggle Twig linting. Are we good to resolve this conversation?

markdorison (approved):

PR #35, opened 2021-09-28
Adds setup:codespaces command

walangitan:
## Description
Adds a `setup:codespaces` command that does the following:

1. Symlink `/var/www/html` to the `/web` directory on a site.
2. Start apache and forward port 80.
3. Download the database.
4. Import the database.
5. Build the theme.
6. Clear the cache.

## Motivation / Context
https://github.com/ChromaticHQ/chromatichq.com/pull/3472/ has issues when executing other robo commands from another robofile, returning a `$result` object instead of a string that we need for our database imports.

Testing in the https://github.com/ChromaticHQ/chromatichq.com/pull/3472/ codespace.

## Screenshots
TBD

## Documentation
TBD

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -516,7 +516,7 @@ class DevelopmentModeCommands extends Tasks
     {
         $this->io()->title('Symlinking file system.');
         $docRootDir = Robo::config()->get('drupal_document_root') ?? 'web';
-        $codespaces_directory = getenv('PWD') . '/' .$docRootDir;
+        $codespaces_directory = getenv('PWD') . '/' . $docRootDir;
         if (empty($codespaces_directory)) {
             throw new TaskException($this, 'Codespaces directory is unavailable.');
         }
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -514,7 +514,7 @@ class DevelopmentModeCommands extends Tasks
      */
     public function setupCodespaces()
     {
-        $codespaces_directory = getenv('OLDPWD');
+        $codespaces_directory = getenv('PWD') . '/web';
         if (empty($codespaces_directory)) {
             throw new TaskException($this, 'Codespaces directory is unavailable.');
         }
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -544,7 +544,16 @@ class DevelopmentModeCommands extends Tasks
         $result = $this->taskExec('composer robo theme:build')->run();
 
         $this->io()->section('Drush deploy.');
-        $result = $this->taskExec('vendor/bin/drush deploy --yes')->run();
+        $result = $this->taskExecStack()
+            ->exec("vendor/bin/drush deploy --yes")
+            // Import the latest configuration again. This includes the latest
+            // configuration_split configuration. Importing this twice ensures that
+            // the latter command enables and disables modules based upon the most up
+            // to date configuration. Additional information and discussion can be
+            // found here:
+            // https://github.com/drush-ops/drush/issues/2449#issuecomment-708655673
+            ->exec("drush config:import --yes")
+            ->run();
         return $result;
     }
 }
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -506,4 +506,39 @@ public function frontendDevDisable($siteDir = 'default', array $opts = ['yes|y'
...
     }
+
+    /**
+     * Setup a site in GitHub Codespaces.
+     *
+     * @throws \Robo\Exception\TaskException
+     */
+    public function setupCodespaces()
+    {
+        $codespaces_directory = getenv('PWD') . '/web';
```
walangitan: This returns the default directory at first, which is  `/workspaces/ {{reponame}}`. I'm appending `/web` here to match CHQ, but some projects use `docroot` still?

adamzimmermann:
```
$docRootDir = Robo::config()->get('drupal_document_root') ?? 'web';
```
Does this line help?

I stole it from elsewhere in the file.

adamzimmermann: Example in the wild:

- https://github.com/ChromaticHQ/imprivata-drupal-8/blob/master/robo.yml#L26

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -506,4 +506,39 @@ public function frontendDevDisable($siteDir = 'default', array $opts = ['yes|y'
...
+        if (empty($codespaces_directory)) {
+            throw new TaskException($this, 'Codespaces directory is unavailable.');
+        }
+        $result = $this->taskExec('rm /var/www/html')->run();
```
adamzimmermann: There are commands to do some of this stuff in a more "robo way" FWIW.

https://robo.li/tasks/Filesystem/

adamzimmermann (commented):
I'm not sure if you are looking for final approval on this, or just a review, so I'll just leave a comment for now.

I'm guessing you want to finish the parent PR while using this as a dev branch and then merge this and tag it for release?

adamzimmermann:
Overall this is looking great so far! 💯

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -506,4 +506,45 @@ public function frontendDevDisable($siteDir = 'default', array $opts = ['yes|y'
...
+        if (empty($dbPath)) {
+            throw new TaskException($this, 'Database download failed.');
+        }
+
+        $this->io()->section('Importing database.');
+        $result = $this->taskExec("zcat $dbPath | mysql -h db -u mariadb -pmariadb mariadb")->run();
+        $result = $this->taskExec('rm')->args($dbPath)->run();
+
+        $this->io()->section('Building theme.');
+        $result = $this->taskExec('composer robo theme:build')->run();
```
markdorison, line 544: This and the db download line above will work for simple sites, but multi-sites will require some special cases. I am OK with merging this assuming we want to address that in a future update.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -506,4 +506,45 @@ public function frontendDevDisable($siteDir = 'default', array $opts = ['yes|y'
...
+
+        $this->io()->section('Clearing Drupal cache.');
+        $result = $this->taskExec('vendor/bin/drush cr')->run();
```
markdorison: Do we want to run just a `cr` or do we need a full `drush deploy` here?

walangitan: I think we do want to use `drush deploy`, since we're importing config twice in lando, we should probably do the same here as well to ensure we're fully up to date?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
walangitan: I can log a ticket about updating this to support multi-sites, as well as some other caveats that may need to be created to support running a multi-site in codespaces.

agarzola (commented):
@walangitan This PR appears to perform all the tasks we need it to complete. However, I did notice in the web VSCode terminal that the process does not appear to exit:

image

@adamzimmermann @markdorison Is there something we need to run here to explicitly exit the process once a task has completed? As far as I can tell, the new `setupCodespaces()` function looks like just any other, so I doubt there is anything else we’re missing here, but I ask just in case.

markdorison:
> @adamzimmermann @markdorison Is there something we need to run here to explicitly exit the process once a task has completed?

No, not that I am aware of. 😕

PR #36, opened 2021-10-04
Exclude node_modules from Twig linting.

adamzimmermann:
## Description

Exclude node_modules from Twig linting.

## Motivation / Context

Once theme dependencies are built locally, the `composer cs-check` command will often fail when it finds malformed Twig templates in external dependencies. This will exclude those directories from Twig linting.

👀 I proposed adding this to the default finder in the Twig linting repo. If my idea is accepted there, this PR should be closed.

- https://github.com/VincentLanglet/Twig-CS-Fixer/issues/40

Required this branch into a repo that previously had this issue, then ran `composer cs-check` and saw it pass with no issues.

## Screenshots

Screen Shot 2021-10-04 at 4 18 19 PM

## Documentation
Maybe?

### DIFF
--- a/.twig-cs-fixer.php
+++ b/.twig-cs-fixer.php
@@ -1,0 +1,9 @@
+<?php
+
+$finder = new TwigCsFixer\File\Finder();
+$finder->exclude('node_modules');
+
+$config = new TwigCsFixer\Config\Config();
+$config->setFinder($finder);
+
+return $config;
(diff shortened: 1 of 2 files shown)


adamzimmermann:
https://github.com/VincentLanglet/Twig-CS-Fixer/issues/40#event-5428483255

Addresses this. Closing.

PR #37, opened 2021-10-05
Widen Robo support now that Drush supports Robo 3.x

markdorison:
## Description
Widen Robo support now that Drush supports Robo 3.x.

## Motivation / Context
https://github.com/drush-ops/drush/commit/3d91cf932

Tests should pass.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "consolidation/robo": "^1.4 || ^2",
+        "consolidation/robo": "^1.4 || ^2 || ^3",
         "squizlabs/php_codesniffer": "^3.5",
         "phpstan/phpstan": "^0.12.77",
         "phpstan/phpstan-deprecation-rules": "^0.12.6",


adamzimmermann (approved):
Approving pending tests passing. ✅ 

I'm curious why we didn't allow the newer version before? If a project also required `drush/drush`, wouldn't composer just sort that out for us?

markdorison:
> I'm curious why we didn't allow the newer version before? If a project also required drush/drush, wouldn't composer just sort that out for us?

@adamzimmermann We don't technically need to require robo at all since we require drush, but there was pushback on removing it from our explicit dependencies since we explicitly used it.

Once we added it to our explicitly listed dependencies, I used the versions that I knew were valid for our uses.

adamzimmermann:
> but there was pushback on removing it from our explicit dependencies since we explicitly used it.

🤚🏼 me haha.

---

> Once we added it to our explicitly listed dependencies, I used the versions that I knew were valid for our uses.

✅

PR #47, opened 2021-11-05
Create Legacy Development Commands for Drupal 7

adamzimmermann:
## Description
Allow commands to continue if an example file is missing.

## Motivation / Context

- If one of these files is missing, the whole command will fail.
- https://github.com/ChromaticHQ/benz-tickets/issues/2065

TK

## Screenshots
TK


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -111,7 +111,7 @@ public function databaseDownload($siteName = 'default')
         $objects = $s3->listObjectsV2($this->s3BucketRequestConfig($siteName));
         $objects = iterator_to_array($objects);
         if (empty($objects)) {
-             throw new TaskException($this, "No database dumps found for '$siteName'.");
+            throw new TaskException($this, "No database dumps found for '$siteName'.");
```
adamzimmermann: My IDE fixed the indentation here.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -435,10 +435,26 @@ public function frontendDevEnable($siteDir = 'default', array $opts = ['yes|y' =
         $this->io()->title('enabling front-end development mode.');
         $this->say("copying settings.local.php and development.services.yml into sites/$siteDir.");
 
-        $result = $this->taskFilesystemStack()
-            ->copy("$this->drupalRoot/sites/example.settings.local.php", $devSettingsPath, true)
-            ->copy("$this->drupalRoot/sites/development.services.yml", $this->devServicesPath, true)
-            ->run();
+        // Copy the example local settings file.
```
adamzimmermann: This is the current state. Do we like how this is handled so far?

command

The next question is how we want to handle D7 in general, as this is really indicative of a larger problem as seen my the next error with the `development.services.yml` file and the other code that attempts to do more D8+ changes.

adamzimmermann (commented):
I applied one bandaid and discovered we needed to keep applying more. I figured we should step back and talk about our larger approach.

Does D7 use a separate branch? Do we default to D8 mode, but have a D7 mode also in the same branch?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: We will create a new D7 specific development command for this use case.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -275,20 +275,19 @@ public function databaseRefreshLando(string $siteName = 'default'): Result
     public function drupalLoginLink($siteDir = 'default', $lando = true): Result
     {
         $this->io()->section("create login link.");
-        $docRootDir = Robo::config()->get('drupal_document_root') ?? 'web';
```
adamzimmermann, line 280: I noticed that we weren't using the class property that already had this. Was there a reason for that?

I can say that this property wasn't set correctly for D7 sites and I figured out what we needed to fix that to allow https://github.com/webflo/drupal-finder to work correctly.

- https://github.com/webflo/drupal-finder/blob/master/src/DrupalFinder.php#L209-L211

We might need to add this for every site that uses this new update. We can put that in the release notes.

src/Robo/Plugin/Commands/LegacyDevelopmentModeCommands.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $this->io()->title('development environment refresh. 🦄✨');
+        $result = $this->taskComposerInstall()->run();
+        // $result = $this->taskExec('lando')->arg('start')->run();
+        // There isn't a great way to call a command in one class from another.
+        // https://github.com/consolidation/Robo/issues/743
+        // For now, it seems like calling robo from within robo works.
+        $result = $this->taskExec("composer robo theme:build $siteName")
+            ->run();
+        $result = $this->frontendDevEnableDrupal7($siteName, ['yes' => true]);
```
adamzimmermann: This is the only line that is different.

adamzimmermann (commented):
I wrestled with this quite a bit trying to come up with an approach that minimized duplication and would be easy to maintain going forward. I'm not in love with the approach, but it was the best approach that came to mind thus far. Open to new ideas and approaches.

adamzimmermann (commented):
@markdorison this is really just what we discussed with the concept of a base class mixed in to help me conceptualize the approach. Now that the approach is sorted out, we could go back to how it was with just two classes, but I think the base class make it a bit more clear what we are doing.

markdorison:
Yeah, I like the base class. We can always iterate from here. This isn't marked as ready for review but FWIW I tested this on CHQ and it worked for me without regressions.

adamzimmermann:
I tested this on D7 and it was working for me too. We can do some more testing next week.

src/Robo/Plugin/Commands/LegacyDevelopmentModeCommands.php
```diff
@@ -0,0 +1,100 @@
...
+
+namespace ChqRobo\Robo\Plugin\Commands;
+
+use Robo\Exception\TaskException;
+use Robo\Result;
+
+/**
+ * Robo commands related to changing development modes for Drupal 7.
+ */
+class LegacyDevelopmentModeCommands extends DevelopmentModeBaseCommands
```
apotek: I wonder whether this should be called `Drupal7DevelopmentModeCommands` rather than `LegacyDevelopmentModeCommands`... because Drupal 8 is now also legacy. It's a minor thing to call out, but I think this should be named specifically for what it is targeting. What do you think @adamzimmermann ?

Other than that, I think this approach is sane. I would privilege being able to discard/change this code easily, make it easily maintainable rather than look for an airtight way to minimize code duplication etc, so I think the way you've factored this makes a ton of sense.

adamzimmermann: I considered that, but putting a version number in a method name seemed weird. I also see your point about D8 being legacy now, so I'm conflicted.

@markdorison any preferences on your end? `Drupal7DevelopmentModeCommands` might be more clear, which is ultimately more important than any weird naming whims on my end. 😂 

I do mention `Drupal 7` in the comment though FWIW. 🤷🏼

markdorison: I would lean towards `Drupal7DevelopmentModeCommands`

adamzimmermann: Done. ✅

adamzimmermann:
This is ready for another (hopefully final) round of review.

src/Robo/Plugin/Commands/Drupal7DevelopmentModeCommands.php
```diff
@@ -0,0 +1,100 @@
...
+     * downloading a database dump, importing it, running deployment commands, disabling front-end caches,
+     * and providing a login link.
+     *
+     * @param string $siteName
+     *   The Drupal site name.
+     *
+     * @return \Robo\Result
+     *   The result of the set of tasks.
+     */
+    public function devRefreshLegacy(string $siteName = 'default'): Result
```
markdorison: Maybe this should be D7 specific?

```suggestion
    public function devRefreshDrupal7(string $siteName = 'default'): Result
```

markdorison (approved):
Tested on CHQ. Working as expected.

adamzimmermann:
Tested on a D7 site and working as expected. 

- Database downloads (`composer robo frontend:dev-enable-drupal7`).
- Front-end dev mode enabled without errors (`composer robo dev:refresh-drupal7`).

@apotek do want to give this one final review, then we can merge this?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -0,0 +1,376 @@
...
+                ->arg('user:login')
+                ->option('--uri', $uri)
+                ->dir("$this->drupalRoot/sites/$siteDir")
+                ->run();
+        }
+        return $this->taskExec('../../../drush')
+            ->arg('user:login')
+            ->dir("$this->drupalRoot/sites/$siteDir")
+            ->run();
+    }
```
apotek, line 162: Honest question: Given our lando and robo usage, is it reasonable to assume drush will always be at this hardcoded path? 

An _optimisation_ here for durability/maintenance would be to have this be $this->drush. But I am overthinking it and this comment should be ignored. It's just a thought.

apotek (approved):
Looking good.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: Great catch. This feels weird. Why is it not calling `drush` in `vendor/bin`?

adamzimmermann: I learned something new with the solution. That `DrupalFinder` class is magic. 🔮 

@markdorison are you still setup to test this easily, or should I?

markdorison: @adamzimmermann I am booked up at the moment. If you can, fire away.

adamzimmermann:
1. This is working with the latest change. 🎉 
2. In my testing I have really stumped myself. 🤯 

### I discovered that subsequent arguments beyond the first one are not passed into the function. So this was always using Lando and was broken when I tried to test it with my local non-Lando setup.

Screen Shot 2021-12-09 at 4 37 14 PM


### So I read [the documentation](https://github.com/consolidation/Robo/blob/3.x/docs/getting-started.md#options) and decided to use option flags instead of arguments. I tried every variation I could think of and nothing seemed to work.


Screen Shot 2021-12-09 at 4 36 11 PM


### I also confirmed that our usage of options in another command are broken too.

Screen Shot 2021-12-09 at 4 41 32 PM


I looked at examples online, confirmed I was reading docs for Robo3 and that we were on Robo3. I even read the [consolidation/annotated-command documentation](https://github.com/consolidation/annotated-command) and nothing fixed it. 

I went ahead and committed my option work here: 29029f5ea0f3bb6916ee24bf824d691411cad5d6

Can someone else take a look at this and spot my mistake? 🙏🏼

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: A couple thoughts, though they may not be relevant.

1. When I look at the database refresh pattern there is one function for using lando, and one for using tugboat. Would it be better to follow that pattern and write two functions here?

2. Or, as long as we are willing to default to using lando, you could either:
  a. check if lando exists and use lando if it does 
  b. `try{ /lando implementation/} catch { try {/non-lando implementation}} finally{/uh oh/}`

adamzimmermann: Did you mean to leave this comment elsewhere? I don't think it is related to this thread.

1. Yes, but I removed that line.
2. I'm not sure what you are getting at here, but I think the answer is no.
3. That's a good question. I would defer to @markdorison since he authored most of this.
4. The thing is Lando might exist and work, but that doesn't mean we want to use it, but perhaps we should just make that assumption and if someone wants a login link they can do it the old school `drush uli` route and then we can just get rid of the non-Lando path in the code.
5. Good question. I'm not sure.

I think answering that helped! Now we just need to see if @markdorison is on board now with ditching Lando support in the login robo command.

apotek: @adamzimmermann I don't know how my comment ended up here. I was trying to respond to a recent commit/question you posted to us regarding how to get the login working in both lando and without lando. I will move my comment down to where it's supposed to be. Sorry

I moved my comment down below your question. I am sorry for the confusion.

apotek:
> Can someone else take a look at this and spot my mistake? 🙏🏼

A couple thoughts, though they may not be relevant.

1. Is the return of `getenv('LANDO_INFO')` truthy enough for PHP?
2. does this script run under the same shell user and environment as you, calling it?
3. When I look at the database refresh pattern there is one function for using lando, and one for using tugboat. Would it be better to follow that pattern and write two functions here?
4. Or, as long as we are willing to default to using lando, you could either:
  a. check if lando exists and use lando if it does, or
  b. `try{ /lando implementation/} catch { try {/non-lando implementation/}} finally{/uh oh/}` 
5. Will it help to think "what if we also end up using ddev" 🙀 ? Or hinder?

Hoping these questions will help you and I am sorry I can't point here and go "there is the problem!" I don't see it either.

markdorison:
@adamzimmermann @apotek If testing this is the thing that uncovered that it's broken for non-Lando users then its clearly not being used much by that group of users. 

If this is working for Lando, then let's merge this and create a follow-up ticket for the broken piece that is not Hitachi/D7 related.

PR #48, opened 2021-11-08
Fixed method exposed as command.

markdorison:
## Description
Fixed method exposed as robo command by switching it from from public to protected.

## Motivation / Context
CleanShot 2021-11-08 at 17 30 53@2x

Tested locally. 

1. Method is no longer exposed as a command.
2. Downstream commands that consume this trait and use this method still work as expected.


walangitan (approved):
Not seeing this appear in the list anymore when I call this branch in a project and run `composer robo list`.

PR #49, opened 2021-11-11
Increase PHPStan level.

markdorison:
## Description
Increased PHPStan level to 6.

## Motivation / Context
Additional static analysis checks.

Tests should pass. No (intended) functional changes.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,6 @@
 parameters:
   customRulesetUsed: true
-  level: 4
+  level: 5
   reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     - message: '#\Call to deprecated method io\(\) of class Robo\\Tasks#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,6 @@
 parameters:
   customRulesetUsed: true
-  level: 1
+  level: 4
   reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     - message: '#\Call to deprecated method io\(\) of class Robo\\Tasks#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,10 +1,11 @@
 parameters:
+  checkMissingIterableValueType: false
   customRulesetUsed: true
-  level: 5
-  reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     - message: '#\Call to deprecated method io\(\) of class Robo\\Tasks#'
       paths:
         - 'src/'
+  level: 6
+  reportUnmatchedIgnoredErrors: true
 includes:
   - vendor/phpstan/phpstan-deprecation-rules/rules.neon
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,6 @@
 parameters:
   customRulesetUsed: true
+  level: 1
   reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     - message: '#\Call to deprecated method io\(\) of class Robo\\Tasks#'
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -16,14 +16,14 @@ class CICommands extends Tasks
     /**
      * Array containing the file extensions PHPCS should check.
      *
-     * @var array
+     * @var string[]
      */
     protected $phpcsCheckExtensions;
 
     /**
      * Array containing paths PHPCS should ignore.
      *
-     * @var array
+     * @var string[]
      */
     protected $phpcsIgnorePaths;
 
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -52,7 +52,7 @@ class CICommands extends Tasks
         $this->phpcsCheckExtensions = Robo::config()->get('phpcs_check_extensions');
         $this->phpcsIgnorePaths = Robo::config()->get('phpcs_ignore_paths');
         $this->customCodePaths = implode(' ', $this->getCustomCodePaths());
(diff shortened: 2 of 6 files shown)


phpstan.neon
```diff
@@ -1,9 +1,11 @@
 parameters:
+  checkMissingIterableValueType: false
   customRulesetUsed: true
-  reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     - message: '#\Call to deprecated method io\(\) of class Robo\\Tasks#'
       paths:
         - 'src/'
+  level: 6
+  reportUnmatchedIgnoredErrors: true
```
apotek, line 9: nice ordering of the params :).

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -52,7 +52,7 @@ public function __construct()
         $this->phpcsCheckExtensions = Robo::config()->get('phpcs_check_extensions');
         $this->phpcsIgnorePaths = Robo::config()->get('phpcs_ignore_paths');
         $this->customCodePaths = implode(' ', $this->getCustomCodePaths());
-        $this->lintTwigFiles = Robo::config()->get('twig_lint_enable', true);
+        $this->lintTwigFiles = Robo::config()->get('twig_lint_enable') ?? true;
```
apotek, line 55: Interesting change from relying on the `get()` method to relying on php itself. I would love to hear your thoughts on the change and why you think it is better to provide the default via php rather than using the `get()` methods provision for that. I have no opinion on the matter.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -139,8 +139,10 @@ public function databaseDownload($siteName = 'default')
      *   Path to the AWS configuration directory.
      * @param string $awsConfigFilePath
      *   Path to the AWS configuration file.
+     *
+     * @return \Robo\Result
      */
-    protected function configureAwsCredentials(string $awsConfigDirPath, string $awsConfigFilePath)
+    protected function configureAwsCredentials(string $awsConfigDirPath, string $awsConfigFilePath): Result
```
apotek, line 145: Question: Why does the docblock need fully qualified class (\Robo\Result) wherease in the function signature the returned class is just Result instead of \Robo\Result. I assume this file has a `use \Robo\Result` or such, which is why the codebase resolves the return type. Is doc syntax not able to resolve the intended class the way the code does?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -337,12 +339,13 @@ protected function landoUri($siteDir): string
...
      * Refresh database on Tugboat.
      *
-     * @return null|\Robo\Result
+     * @return \Robo\Result
      *   The task result.
      */
     public function databaseRefreshTugboat(): Result
     {
         $this->io()->title('refresh tugboat databases.');
+        $result = null;
```
apotek, line 348: `null` is ok. It would be better if Robot provided a way of instantiating an empty \Robo\Result, so the signature could always return a Robo Result instead of `null`.

src/Robo/Plugin/Traits/SitesConfigTrait.php
```diff
@@ -11,12 +11,17 @@
 trait SitesConfigTrait
 {
 
+    /**
+     * Filename for a sites configuration file.
+     *
+     * @var string
+     */
```
apotek, line 18: +1

apotek (approved):
Good changes. Nice clean work.

src/Robo/Plugin/Commands/CICommands.php
markdorison: I think the issue was that the second (default) parameter was typed as a `string` and it didn't like that I was specifying a `bool`, but I _wanted_ a `bool`, so this was my solution.

apotek: Makes perfect sense then. Basically we are covering for a not yet up to spec implementation in Robo.

adamzimmermann: 💯 on the question and the answer. I'm sold.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: > Why does the docblock need fully qualified class (\Robo\Result) wherease in the function signature the returned class is just Result instead of \Robo\Result.

I've wondered this too when I first started getting my hands dirty with OO and D8. FWIW this is just how I've always seen it done and what documentation always says to do. 

> Is doc syntax not able to resolve the intended class the way the code does?

That is my best guess, but I don't have a good answer either.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
adamzimmermann: I like this idea. ☝🏼 

If this idea doesn't pan out, does this mean that our docblock should be `@return \Robo\Result|null`?

src/Robo/Plugin/Traits/SitesConfigTrait.php
```diff
@@ -11,12 +11,17 @@
 trait SitesConfigTrait
 {
 
+    /**
+     * Filename for a sites configuration file.
```
adamzimmermann:
```suggestion
     * Filename for a site's configuration file.
```

adamzimmermann (approved):
A few questions, but overall I'm sold. Excited to see this being integrated.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: I like it too. PRs welcome! 😉

PR #52, opened 2021-12-16
Rename project to usher.

markdorison:
## Description
Renam project to usher.

## Motivation / Context
Give the project a more unique name.

Tested locally. Commands are all still available.

## Deployment Instructions
Once deployed:

- [x] Rename Github repo.
- [x] Resubmit the package to packagist using the new name.
- [x] Mark the old package as "Abandoned" on packagist, and use the new name in the form so that people get pointed to it when they install with the old name.
- [ ] Update Chromatic projects that reference the package (even if not required, good for clarity).

## Documentation
Updated.


adamzimmermann (approved):
I don't understand the name, but the code changes look good.

PR #57, opened 2022-02-21
Added requirement for dealerdirect/phpcodesniffer-composer-installer.

markdorison:
## Description
Added requirement for dealerdirect/phpcodesniffer-composer-installer.

## Motivation / Context
This was often/always included in downstream repos. Moving it here as Usher assumes its presence. 

Tested locally.


apotek (approved):
Composer changes only. Look good.

PR #58, opened 2022-03-02
Update README to use chromatic/usher

mmatsoo:
## Description
The instruction currently reads `composer require chromatichq/usher`, but it should be `composer require chromatic/usher`, without the "hq" in "chromatichq".

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ downstream repos.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


adamzimmermann (approved):
👏🏼👏🏼

markdorison (approved):

PR #63, opened 2022-03-04
Add dependency on phpstan-extension-installer.

markdorison:
## Description
* Add dependency on [phpstan-extension-installer](https://github.com/phpstan/extension-installer).
* Remove explicit include that is no longer needed.

## Motivation / Context
CleanShot 2022-03-04 at 11 38 00@2x

Tests should pass.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -7,5 +7,3 @@ parameters:
         - 'src/'
   level: 6
   reportUnmatchedIgnoredErrors: true
-includes:
-  - vendor/phpstan/phpstan-deprecation-rules/rules.neon
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "drush/drush": "^8 || ^10 || ^11",
         "webflo/drupal-finder": "^1.2",
         "vincentlanglet/twig-cs-fixer": "^0.4.1",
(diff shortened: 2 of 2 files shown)


phpstan.neon
```diff
@@ -7,5 +7,3 @@ parameters:
         - 'src/'
   level: 6
   reportUnmatchedIgnoredErrors: true
-includes:
-  - vendor/phpstan/phpstan-deprecation-rules/rules.neon
```
adamzimmermann, line 11: We also have `vendor/mglaman/phpstan-drupal/extension.neon` in CHQ. Do we need to add that extension to Usher?

adamzimmermann (approved):

phpstan.neon
markdorison: I think we would leave that for the Drupal sites to install themselves, especially if `phpstan-extension-installer` is going to install/activate it automatically.

In fact, I just tested it and it will result in an error on codebases that do not have Drupal installed (like this one).

`phpstan-drupal/drupal-autoloader.php: Unable to detect Drupal at`

adamzimmermann: Good call. Let's add that to the release notes for the next Usher release and try to remember to get this added to the other repos (where applicable).

PR #64, opened 2022-03-07
Implemented phpstan-strict-rules

markdorison:
## Description
* Added `phpstan-strict-rules` as dev dependency (so it is not installed in downstream repos).
* Fixed issues reported by strict rules.

## Motivation / Context
Testing stricter rules on this small codebase to determine how worthwhile we think they are.

* Tests should pass.
* Installed with a downstream repo to ensure it does not cause unintended issues. Worked as expected.

### Before Fixes
CleanShot 2022-03-10 at 13 38 33@2x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
         "webflo/drupal-finder": "^1.2",
         "vincentlanglet/twig-cs-fixer": "^0.4.1",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
-        "phpstan/extension-installer": "^1.1"
+        "phpstan/extension-installer": "^1.1",
+        "phpstan/phpstan-strict-rules": "^1.1"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -19,8 +19,7 @@
         "webflo/drupal-finder": "^1.2",
         "vincentlanglet/twig-cs-fixer": "^0.4.1",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
-        "phpstan/extension-installer": "^1.1",
-        "phpstan/phpstan-strict-rules": "^1.1"
+        "phpstan/extension-installer": "^1.1"
     },
     "autoload": {
         "psr-4": {
@@ -42,5 +41,8 @@
             "dealerdirect/phpcodesniffer-composer-installer": true,
             "phpstan/extension-installer": true
         }
+    },
+    "require-dev": {
+        "phpstan/phpstan-strict-rules": "^1.1"
     }
 }
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -142,7 +142,7 @@ class CICommands extends Tasks
      */
     protected function getCustomCodePaths(): array
     {
-        if (!$customCodePaths = Robo::config()->get('custom_code_paths')) {
+        if (!is_array($customCodePaths = Robo::config()->get('custom_code_paths'))) {
             throw new TaskException($this, 'Expected Robo configuration not present: custom_code_paths');
         }
         return $customCodePaths;
@@ -158,7 +158,7 @@ class CICommands extends Tasks
      */
     protected function getCodingStandards(): array
     {
-        if (!$phpcsStandards = Robo::config()->get('phpcs_standards')) {
+        if (!is_array($phpcsStandards = Robo::config()->get('phpcs_standards'))) {
             throw new TaskException($this, 'Expected Robo configuration not present: phpcs_standards');
         }
         return $phpcsStandards;
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -120,7 +120,10 @@ class CICommands extends Tasks
     }
 
     /**
-     * Command to fix coding standards where possible.
+     * Determine whether to check or fix coding standard issues and run as requested.
+     *
+     * @param bool $applyFixes
+     *   FALSE if phpcs should be run, else phpcbf.
      *
      * @return \Robo\Result
      *   The result of the set of tasks.
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -13,16 +13,16 @@ use Robo\Tasks;
 class CICommands extends Tasks
 {
     /**
-     * Array containing the file extensions PHPCS should check.
+     * A comma separated list of the file extensions PHPCS should check.
      *
-     * @var string[]
+     * @var string
      */
     protected $phpcsCheckExtensions;
 
     /**
-     * Array containing paths PHPCS should ignore.
+     * A comma separated list of the paths PHPCS should ignore.
      *
-     * @var string[]
+     * @var string
      */
     protected $phpcsIgnorePaths;
 
@@ -32,6 +32,13 @@ class CICommands extends Tasks
      * @var string
      */
     protected $customCodePaths;
+
+    /**
+     * A comma separated list of PHPCS standards.
+     *
+     * @var string
+     */
+    protected $phpcsStandards;
 
     /**
      * Boolean indicating whether Twig files should be linted.
@@ -48,9 +55,10 @@ class CICommands extends Tasks
         // Treat this command like bash -e and exit as soon as there's a failure.
         $this->stopOnFail();
 
-        $this->phpcsCheckExtensions = Robo::config()->get('phpcs_check_extensions');
-        $this->phpcsIgnorePaths = Robo::config()->get('phpcs_ignore_paths');
-        $this->customCodePaths = implode(' ', $this->getCustomCodePaths());
+        $this->phpcsCheckExtensions = implode(',', Robo::config()->get('phpcs_check_extensions'));
+        $this->phpcsIgnorePaths = implode(',', Robo::config()->get('phpcs_ignore_paths'));
+        $this->customCodePaths = implode(' ', $this->getConfigurationValues('custom_code_paths'));
+        $this->phpcsStandards = implode(',', $this->getConfigurationValues('phpcs_standards'));
         $this->lintTwigFiles = Robo::config()->get('twig_lint_enable') ?? true;
     }
 
@@ -95,17 +103,7 @@ class CICommands extends Tasks
      */
     public function jobCheckCodingStandards(): Result
     {
-        $standards = implode(',', $this->getCodingStandards());
-        $extensions = implode(',', $this->phpcsCheckExtensions);
-        $ignorePaths = implode(',', $this->phpcsIgnorePaths);
-        /** @var \Robo\Task\CommandStack $stack */
-        $stack = $this->taskExecStack()->stopOnFail();
-        $stack->exec("vendor/bin/phpcs --standard=$standards --extensions=$extensions \
-            --ignore=$ignorePaths $this->customCodePaths");
-        if ($this->lintTwigFiles) {
-            $stack->exec("vendor/bin/twig-cs-fixer lint $this->customCodePaths");
-        }
-        return $stack->run();
+        return $this->jobRunCodingStandards(false);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -142,7 +142,7 @@ public function jobFixCodingStandards(): Result
      */
     protected function getCustomCodePaths(): array
     {
-        if (!$customCodePaths = Robo::config()->get('custom_code_paths')) {
+        if (!is_array($customCodePaths = Robo::config()->get('custom_code_paths'))) {
```
apotek: Do we control `Robo::config()->get()`? If so, I would want to change the function signature to return something false if there is not an array being returned, and keep the old code. Otherwise, every time you call that function, you have to know more about its internals. 

What if you make `get()` _always_ return an array, but an empty one if there are no paths. Make the function more strict and make the code that calls the function work less.

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -158,7 +158,7 @@ protected function getCustomCodePaths(): array
      */
     protected function getCodingStandards(): array
     {
-        if (!$phpcsStandards = Robo::config()->get('phpcs_standards')) {
+        if (!is_array($phpcsStandards = Robo::config()->get('phpcs_standards'))) {
```
apotek: same as above.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -106,7 +106,7 @@ public function databaseRefreshTugboat(): Result
                 // @todo: Should we run a site-install by default?
                 continue;
             }
-            if (empty($dbPath)) {
+            if (!is_string($dbPath) || strlen($dbPath) == 0) {
```
apotek, line 109: Again, I would make `$this->databaseDownload($siteName);` do this work for you so you can simply do `if $dbpath` later.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -253,7 +253,9 @@ protected function landoUri($siteDir): string
                 $domain = array_pop($siteDomains);
                 return "http://$domain";
             }
-        } elseif ($uri = $landoCfg['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI'] ?? null) {
+        } elseif (
+            is_string($uri = $landoCfg['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI'])
```
apotek: Do you also need this to be a non-empty string?

src/Robo/Plugin/Commands/ValidateConfigCommands.php
```diff
@@ -50,7 +50,7 @@ public function validateDrupalConfig($siteDir = 'default'): void
             ->run();
         $drushOutput = trim($result->getOutputData());
         $configJson = json_decode($drushOutput);
-        if (!empty($configJson)) {
+        if (!is_null($configJson)) {
```
apotek, line 53: I am tempted to suggest a `try{}catch{}` around `json_decode()` unless you are guaranteed that `getOutputData()` will return json-decodable output no matter what happens in the task call.  Maybe getOutputData already does that and guarantees JSON in all cases, in which case, please ignore the above.

_But_ if there is a contract  as to what getOutputData will return, I would assume that one of the things in that output data json blob is going to be a success/error state. If that is the case, it seems it would be better to test that value. ie `if ($configJson->status == 0)`

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
```diff
@@ -49,7 +49,7 @@ public function databaseDownload(string $siteName = 'default')
         ]);
         $objects = $s3->listObjectsV2($this->s3BucketRequestConfig($siteName));
         $objects = iterator_to_array($objects);
-        if (empty($objects)) {
+        if (count($objects) == 0) {
```
apotek, line 52: Again, it looks like we are adding robustness in the calling code checks, rather than to the return of the function. If we don't control the signature of `$s3->listObjectsV2` then this is fine (though I would be tempted then to wrap/extend that method to provide that robustness). Try to get to `if ($objects) {` and force the upstream function to be factored in such a way that guarantees a known type of response.

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
```diff
@@ -141,7 +141,7 @@ protected function s3BucketRequestConfig(string $siteName): array
      */
     protected function s3BucketForSite(string $siteName): string
     {
-        if (!$bucket = $this->getConfig('database_s3_bucket', $siteName)) {
+        if (!is_string($bucket = $this->getConfig('database_s3_bucket', $siteName))) {
```
apotek, line 144: Same concern there. The previous line is better, if you can get getConfig() to _behave_.

apotek (commented):
In general it feels like our functions are not strict enough in terms of what you can expect to get back from them, which is forcing you to second guess return values. I feel the solution is to make the function signatures more strict. 

But I also realize we might not have the luxury of time, so I am not going to block these changes, just add my comment. The code looks sane otherwise.

markdorison:
> In general it feels like our functions are not strict enough in terms of what you can expect to get back from them, which is forcing you to second guess return values. I feel the solution is to make the function signatures more strict.

@apotek Maybe I am being too defensive and refusing to trust that the code I am calling won't change unexpectedly? For many of the items you flagged we do not control the methods. I will note those inline.

src/Robo/Plugin/Commands/CICommands.php
markdorison: We do not control [Robo](https://github.com/consolidation/robo), but we do control the [configuration files](https://github.com/ChromaticHQ/usher/blob/2.x/robo.yml) that it is going to be reading. Am I being too defensive? The whole purpose of this method is to validate that we get an array back but if I just called `Robo::config()->get('custom_code_paths'))` [here](https://github.com/ChromaticHQ/usher/blob/phpstan-strict/src/Robo/Plugin/Commands/CICommands.php#L53) we would still get an exception for passing a non-array to `implode()` (I checked).

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: Good catch. Will update.

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
markdorison: We do not control this. It is an AWS PHP library that we are using. Again, maybe I am being too defensive here.

composer.json
```diff
@@ -41,5 +41,8 @@
             "dealerdirect/phpcodesniffer-composer-installer": true,
             "phpstan/extension-installer": true
         }
+    },
+    "require-dev": {
+        "phpstan/phpstan-strict-rules": "^1.1"
```
adamzimmermann, line 46: > Installed with a downstream repo to ensure it does not cause unintended issues. Worked as expected.

I don't see `phpstan/phpstan-strict-rules` in any of our other repos.

Are we adding a new ruleset that will be automatically included and enforced by phpstan in any repo that includes this latest version of `usher`?

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
adamzimmermann: What is the benefit of this change? i.e. what was the linter saying was the issue?

I'm more curious than anything, as I'm also a frequent user of `empty` checks.

adamzimmermann (approved):
One big picture question, but overall I like these changes. ✅ 

With all of these type checks and return types it won't be long until it feels like we are programming in Java, but in a good way. 😂

src/Robo/Plugin/Commands/CICommands.php
apotek: Ok. All my comments assume control of the called function, so they can be ignored. It's too bad the returns are vague.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: We don't control the function so this added carefulness I guess is warranted.

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
apotek: > We do not control this. It is an AWS PHP library that we are using. Again, maybe I am being too defensive here.

I don't feel you are being defensive :). And if these methods don't guarantee a certain type of return value, then you are doing the right thing.

composer.json
markdorison: No, exactly the opposite. By putting this in dev dependencies it will ensure that it is _not_ included in downstream repos.

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
markdorison: > What is the benefit of this change? i.e. what was the linter saying was the issue?
> 
> I'm more curious than anything, as I'm also a frequent user of empty checks.
> 

@adamzimmermann [phpstan-strict does not like empty()](https://github.com/phpstan/phpstan-strict-rules):

> Disallow empty() - it's a very loose comparison (see [manual](https://php.net/empty)), it's recommended to use more strict one.

src/Robo/Plugin/Commands/CICommands.php
markdorison: > It's too bad the returns are vague.

With my testing I think the returns are actually _not vague_:

> The whole purpose of this method is to validate that we get an array back but if I just called Robo::config()->get('custom_code_paths')) [here](https://github.com/ChromaticHQ/usher/blob/phpstan-strict/src/Robo/Plugin/Commands/CICommands.php#L53) we would still get an exception for passing a non-array to implode() (I checked).

I am wondering if we can remove this defensiveness because these calls will already return exceptions if the value is not present. 🤔

markdorison: After further testing, I think the error messages we are able to provide with these extra checks are more descriptive and helpful since we know more specifically what we are expecting so I would advocate for leaving them.

markdorison:
@apotek @adamzimmermann This now includes some refactoring that is not limited to the "strict rules" implementation, but you two got me all gassed up so here we are! 😂

composer.json
adamzimmermann: Ah, I didn't think about that feature/angle of using dev dependencies before. The more you know.

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -13,16 +13,16 @@
 class CICommands extends Tasks
 {
     /**
-     * Array containing the file extensions PHPCS should check.
+     * A comma separated list of the file extensions PHPCS should check.
```
adamzimmermann, line 16: 👏🏼

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -118,34 +116,31 @@ public function jobCheckCodingStandards(): Result
...
-     *   An array containing application custom code paths.
+     * @param bool $applyFixes
+     *   FALSE if phpcs should be run, else phpcbf.
      *
-     * @throws \Robo\Exception\TaskException
+     * @return \Robo\Result
+     *   The result of the set of tasks.
      */
-    protected function getCustomCodePaths(): array
+    protected function jobRunCodingStandards(bool $applyFixes = false): Result
```
adamzimmermann, line 131: I'm on the fence about this consolidation. I see pros and cons, but not enough either way to take a stand right now.

So 🎉 I guess? haha.

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -156,11 +151,11 @@ protected function getCustomCodePaths(): array
...
-    protected function getCodingStandards(): array
+    protected function getConfigurationValues(string $key): array
     {
-        if (!$phpcsStandards = Robo::config()->get('phpcs_standards')) {
-            throw new TaskException($this, 'Expected Robo configuration not present: phpcs_standards');
+        if (!is_array($configurationValues = Robo::config()->get($key))) {
+            throw new TaskException($this, "Expected Robo configuration not or malfomed present: $key");
         }
-        return $phpcsStandards;
+        return $configurationValues;
```
adamzimmermann, line 159: I like this rename. 💯

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -253,9 +253,9 @@ protected function landoUri($siteDir): string
                 $domain = array_pop($siteDomains);
                 return "http://$domain";
             }
-        } elseif ($uri = $landoCfg['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI'] ?? null) {
+        } elseif (isset($landoCfg['services']['appserver']['overrides']['environment']['DRUSH_OPTIONS_URI'])) {
```
adamzimmermann, line 256: Just confirming that we wanted to change the logic here. As we returned `null` before without getting to the `else` below. 

I think this is better though FWIW.

src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
adamzimmermann: Ah, see that's what I like about it in some circumstances.

adamzimmermann (approved):

composer.json
apotek: I am so confused. I think you both might be saying the same thing? 

My primitive (non)understanding is that by putting it into require-dev, it won't be automatically included downstream, but is _available_ downstream for a dev install, and isn't that what @adamzimmermann is saying?

🤔

adamzimmermann: > but is available downstream for a dev install, and isn't that what @adamzimmermann is saying?

I wasn't saying that, but that is a good point. I didn't know the answer to this, so I did some research.

> NB composer install will only download and install the development dependencies of your application or package and not the development dependencies of dependencies. So if you have a dependency on "Package A" and "Package A" has a development dependency on "Package B" only "Package A" will be downloaded and installed in your project.

— https://dev.to/timoschinkel/good-practices-when-working-with-composer-5a1c

This also lines up with what @markdorison is implying, so that's good.

apotek: Thank you for digging in there @adamzimmermann . That quote aligns (I am relieved to see) with my thinking but I guess what is a surprise is that if I do a `composer install --dev` I guess I had always assumed that my dependency dev requirements would be included (ie, that the require-dev sections would inherit). But I guess I am wrong .

adamzimmermann: I had honestly just never thought about it (or forgot if I did), which kind of surprises me, as I've thought way too much about composer, but I think it makes sense.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: > As we returned null before without getting to the else below.

That is not my reading of it. If it was not set, the null coalesce would have returned NULL, but then the `elseif` would have failed the check and it would have moved on to the `else`.

adamzimmermann: 👍🏼

PR #70, opened 2022-04-07
`composer robo uli` on a site with one` .lando.yml` and Front-End and Back-End living in seperate sub-directories fix

### LINKED ISSUE #69: `composer robo uli` on a site with one .lando.yml and Front-End and Back-End living in seperate sub-directories goes CLUNK
## Description

Encountered this on the Poets site, which has one `.lando.yml` in the root-directory, a Drupal Back-end in a sub-directory of this root-directory and the Front-End living in another sub-directory.

Using Usher to setup the Lando dev-environment. Since there's only one `composer.json` (in the Drupal sub-dir) the `composer robo` magic takes place there.

The `composer robo uli` however expects the `.lando.yml` (as is the "normal" use-case) to live in one-up from the `DRUPAL_ROOT ` dir.

In this particular case `.lando.yml` lives two directories above the `DRUPAL_ROOT`.

This leads to a CLUNCK [here](https://github.com/ChromaticHQ/usher/blob/2.x/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php#L236) when ` Yaml::parseFile($landoConfigPath);` throws a `ParseException`.

```
create login link.
------------------

 [ExecStack] Done in 06:06
 [error]  File "D:\htdocs\poets\poets-api/web/../.lando.yml" does not exist.
Script robo --ansi handling the robo event returned with error code 1
```

## Proposed Solution

For me and my use case, the Quick-N-Dirty catch-exception-set-$landoConfigPath-to-one-dir-up-and-retry works (see linked PR).
There are endless better and more robust options I can think of, but as a starting point, I'm using that version and submit PR.

## Alternatives Considered

Getting someone with an actual brain looking at it instead for entering the slippery slope of QnD

Boegie:
## Description
The Quick and also Dirty way to solve #69 for Poets.
Please tell me to make this more generic/better.

## Motivation / Context
#69 should have all the motivational and contextual information your heart may desire.

Result with current Usher release on the Poets Lando:
image

Applied the latest commit on my "UsherLite" clone in Poets.

Result after:
image

Also applied the changes to CHQ.com:

Result before:
image

Result after:
image

**NOTE**
The exception thrown on CHQ.com happens with and without the changes. Opened a new issue for that here: #73 .
The line `>  Found Lando config at D:\htdocs\chromatichq.com/web/../.lando.yml.` in the second screenshot proves that the changes also work for the "default" setup of one directory for Front- and Back-End

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -232,8 +232,15 @@ class DevelopmentModeBaseCommands extends Tasks
      */
     protected function landoUri($siteDir): string
     {
-        $landoConfigPath = "$this->drupalRoot/../.lando.yml";
-        $landoCfg = Yaml::parseFile($landoConfigPath);
+        try {
+            $landoCfg = Yaml::parseFile($landoConfigPath);
+        }
+        catch (ParseException $exception) {
+            // This site could have a Front- and Back-end site in different
+            // sub-directories with a the lando.yml in the root directory.
+            $landoConfigPath = "$this->drupalRoot/../../.lando.yml";
+            $landoCfg = Yaml::parseFile($landoConfigPath);
+        }
         // First, check for multisite proxy configuration.
         if (isset($landoCfg['proxy']['appserver'])) {
             if ($siteDir === 'default') {
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -248,7 +248,7 @@ class DevelopmentModeBaseCommands extends Tasks
                 $this->say("Unable to load Lando config from $landoConfigPath.");
             }
             // Break out of the loop if valid configuration was found.
-            if ($landoCfg) {
+            if ($landoCfg === true) {
                 $this->say("Found Lando config at $landoConfigPath.");
                 break;
             }
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -234,8 +234,7 @@ class DevelopmentModeBaseCommands extends Tasks
     {
         try {
             $landoCfg = Yaml::parseFile($landoConfigPath);
-        }
-        catch (ParseException $exception) {
+        } catch (ParseException $exception) {
             // This site could have a Front- and Back-end site in different
             // sub-directories with a the lando.yml in the root directory.
             $landoConfigPath = "$this->drupalRoot/../../.lando.yml";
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -232,8 +233,15 @@ protected function devRefreshDrupal(string $siteName = 'default'): Result
      */
     protected function landoUri($siteDir): string
     {
-        $landoConfigPath = "$this->drupalRoot/../.lando.yml";
-        $landoCfg = Yaml::parseFile($landoConfigPath);
+        try {
```
adamzimmermann: @Boegie what about something like this?

```
// The site could have a front-end and back-end site in different
// subdirectories with the lando.yml in the root directory.
$landoConfigPaths = [
  "$this->drupalRoot/../.lando.yml",
  "$this->drupalRoot/../../.lando.yml",
];
foreach ($landoConfigPaths as $landoConfigPath) {
  try {
    $landoCfg = Yaml::parseFile($landoConfigPath);
  } catch (ParseException $exception) {
    $this->say("Unable to load Lando config from $landoConfigPath.");
  }
  // Break out of the loop if valid configuration was found.
  if ($landoCfg) {
    break;
  }
}
```
This untested and can probably be improved, but hopefully it helps.

adamzimmermann (commented):
Thanks for submitting a fix. 🙏🏼 

I added a suggestion/alteration to your idea. Let me know if that works. 🤞🏼

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
Boegie: Like it, changed it a little

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -232,8 +233,26 @@ protected function devRefreshDrupal(string $siteName = 'default'): Result
...
+            } catch (ParseException $exception) {
+                $this->say("Unable to load Lando config from $landoConfigPath.");
+            }
+            // Break out of the loop if valid configuration was found.
+            if ($landoCfg === true) {
+                $this->say("Found Lando config at $landoConfigPath.");
+                break;
+            }
+        }
         // First, check for multisite proxy configuration.
```
adamzimmermann, line 262:
```suggestion
        if (!$landoCfg) {
             // Abort the process and let the user know what is up.
        }
        // First, check for multisite proxy configuration.
```

adamzimmermann (approved):
I like your adjustments. 💯 

Can we add a bit more error handling to make this even better?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
Boegie: Liked it, did it my way (hit song guaranteed)

markdorison (approved):

adamzimmermann (approved):
I love the attention to detail. 💯

PR #87, opened 2022-08-18
Add cache clear after database import.

adamzimmermann:
## Description
Runs a cache clear after importing the database.

## Motivation / Context

We are exploring trying to reduce the need to run cache rebuilds on Tugboat previews built from the base preview. Storing a base preview with no cache would aid in that. This is an attempt to do that.

https://github.com/ChromaticHQ/benz-tickets/issues/2803

TK


markdorison (approved):

agarzola (approved):
I know this is still a Draft, but I think it’s a good idea regardless of what we do to speed up child previews.

adamzimmermann:
@agarzola without the other changes, this wouldn't speed up child previews. In fact it would slow them down for the first page load, but I think we could argue that it should be done for other reasons.

agarzola:
> @agarzola without the other changes, this wouldn't speed up child previews. In fact it would slow them down for the first page load, but I think we could argue that it should be done for other reasons.

Good point!

PR #89, opened 2022-10-06
Add PHP version linting using phpcompatibility/php-compatibility.

adamzimmermann:
## Description
Adds a standardized to linting for PHP version specific issues.

- https://github.com/PHPCompatibility/PHPCompatibility

⚠️  Should this be a separate command with separate config to allow linting of contrib modules too?

## Motivation / Context
As we update to newer versions of PHP we want to standardize the linting process.

Run `composer robo cs-check` and confirm there are no issues or only valid issues are flagged.

## Screenshots

test

## Documentation
I have added to the example file.


apotek (approved):
LGTM! Nice initiative. This will certainly come in handy.

goranmiric (approved):

PR #90, opened 2022-10-27
Add D7 compatibility for databaseRefreshTugboat().

markdorison:
## Description
Add D7 compatibility for databaseRefreshTugboat().

What do you think of this approach? We could add a separate command, but given its only the [cache clear](https://github.com/ChromaticHQ/usher/commit/4d02216282412f8133de6accee4152c87ffe9b5b) that differs, I opted to wrap it in a conditional.

## Motivation / Context
https://chromatic.slack.com/archives/C02K4QA7V/p1666880292105249?thread_ts=1666796876.984279&cid=C02K4QA7V

Test against this branch.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -92,10 +92,13 @@ class DevelopmentModeBaseCommands extends Tasks
     /**
      * Refresh database on Tugboat.
      *
+     * @param bool $drupal7
+     *   TRUE if running against a Drupal 7 site.
+     *
      * @return \Robo\Result
      *   The result of the set of tasks.
      */
-    public function databaseRefreshTugboat(): Result
+    public function databaseRefreshTugboat(bool $drupal7 = false): Result
     {
         $this->io()->title('refresh tugboat databases.');
         $result = null;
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -124,10 +127,17 @@ public function databaseRefreshTugboat(): Result
                 ->run();
             $result = $this->taskExec('rm')->args($dbPath)->run();
 
-            $result = $this->taskExec("$this->vendorDirectory/bin/drush")
-                ->arg('cache:rebuild')
-                ->dir("$this->drupalRoot/sites/$siteName")
-                ->run();
+            if ($drupal7) {
```
mmatsoo, line 130: @markdorison Should the if/else logic be reversed? Here it says use `drush cr` if we're on D7, which should be `drush cc all`.

mmatsoo: Another issue is `$siteName` because that depends on the key in `.sites.config.yml` and that doesn't necessarily match the site folder names. (E.g. `intuit` vs `intuitbenefits.com`)

markdorison: @mmatsoo Let's hop on a quick zoom. It might be better to just create a custom command for this scenario.

markdorison:
Closing this PR in favor of @mmatsoo pinning to an earlier version.

PR #92, opened 2022-11-02
Notify Slack when Deployment to a Tugboat Base Preview Fails

adamzimmermann:
## Description
Adds a Slack hook integration that notifies Slack when a Tugboat base preview fails to deploy.

## Motivation / Context
When a base preview fails to build and no one knows about it, it disrupts our workflow a lot.

This is intended to help enable proactive problem solving vs reactive problem solving.

- We will need to integrate this with a repo and report back here.
- https://github.com/ChromaticHQ/benz-platform/pull/4349/files

## Screenshots

Screenshot 2022-12-07 at 9 43 45 PM

## Documentation
Included. ✅


src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +84,58 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
...
+            return;
+        }
+        // Build various variables and URLs for the Slack message.
+        $dashboard_url = sprintf(
+            'https://dashboard.tugboatqa.com/%s',
+            getenv('TUGBOAT_PREVIEW_ID'),
+        );
+        // Send the Slack webhook call.
+        $client = new Client();
+        $client->post($slack_webhook_url, [
```
adamzimmermann: I guessed at the structure of this request. This might need to be fixed.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +84,58 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
...
+     * @see https://docs.tugboatqa.com/starter-configs/code-snippets/slack-integration/
+     */
+    protected function notifySlackOnFailedBasePreviewBuild(Result $result): void
+    {
+        // Confirm we are in a Tugboat environment.
+        if (getenv('TUGBOAT_PREVIEW_ID') === false) {
+            return;
+        }
+        // Determine if we are building a base preview.
+        // @todo Remove the testing override condition at the end before merging.
```
adamzimmermann: 🚧 Remove this testing override before merging. ⚠️

composer.json
```diff
@@ -20,7 +20,8 @@
         "vincentlanglet/twig-cs-fixer": "^0.4.1 || ^0.5",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
         "phpstan/extension-installer": "^1.1",
-        "phpcompatibility/php-compatibility": "^9.3"
+        "phpcompatibility/php-compatibility": "^9.3",
+        "guzzlehttp/guzzle": "^6.0"
```
adamzimmermann: I wanted to make this more generic, but wasn't sure how. Ideas are welcome.

src/Robo/Plugin/Traits/ResultCheckTrait.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @param \Robo\Result $result
+     *   The primary result to check within.
+     *
+     * @return bool
+     *   Boolean indicating if all tasks within a result were successful.
+     *
+     * @throws \Robo\Exception\TaskException
+     */
+    public function resultTasksSuccessful(Result $result): bool
```
adamzimmermann: We used to use this before we parallelized tasks. When you run `$result->accumulate()` it requires something like this to really check the result of each individual result within the result. I don't know why Robo doesn't offer a method to do this already. Perhaps this could be a contrib opportunity.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +85,44 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
...
+        if (getenv('TUGBOAT_PREVIEW_ID') !== getenv('TUGBOAT_BASE_PREVIEW_ID') || getenv('TUGBOAT_PREVIEW_ID') !== '') {
+            return;
+        }
+        // If everything went well there is nothing to do.
+        if ($this->resultTasksSuccessful($result)) {
```
adamzimmermann: We need to force a failure to get past this too.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +85,44 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
             // https://github.com/drush-ops/drush/issues/2449#issuecomment-708655673
             ->exec("$appDirPath/vendor/bin/drush config:import --yes")
             ->run();
+        // @todo Do we want/need this option if we can just check for environment variables?
+        // @todo This could be useful if a site wanted to opt out of this functionality.
```
mmatsoo: @adamzimmermann I'm a big fan of opting out of things! Could it be as simple as adding an additional key/value in the `$opts` array,  defaulting to `'optout' => false`?

adamzimmermann: Isn't that what I'm doing with the `tugboat` option essentially? I'm confused. 🤔 

We could rename it to `slack` or something like that if we want.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +85,46 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
...
+     * Notify Slack if a base preview build failed.
+     *
+     * @param \Robo\Result $result
+     *   The result of the task to check.
+     *
+     * @see https://docs.tugboatqa.com/starter-configs/code-snippets/slack-integration/
+     */
+    protected function notifySlackOnFailedBasePreviewBuild(Result $result): void
+    {
+        // @todo Remove this.
```
adamzimmermann: 🚧 Remove this before merging. 🚧

src/Robo/Plugin/Traits/ResultCheckTrait.php
adamzimmermann: Evidently this approach doesn't play nice with parallel task results so I removed it.

src/Robo/Plugin/Commands/DeployCommands.php
apotek: Reminder that not everyone can easily set their environment variables. So much depends on their infrastructure. Opts + env variables is the standard for this reason, so people can either use environment variables if available to them, or opts, if not.

mmatsoo: @adamzimmermann I guess I misunderstood the comments or reacted to `rename this option then to be more generic` and thought about things that needn't be Tugboat-specific. Like a battery of tests I can magically skip with `optout => true`!

adamzimmermann: @mmatsoo I'm not 100% sure I followed your response (sorry!), but are you on board with the code I have in place?

mmatsoo: @adamzimmermann Yes, I'm on board!

Perhaps this pseudo-code will better demonstrate what I was trying to explain about having an additional argument.

```php
        if ($opts['tugboat']) {
            $this->notifySlackOnFailedBasePreviewBuild($result);
        }
        // Allow early optout of additional tests or notifications.
        if ($opts['optout']) {
          return $result;
        }
        // Additional tests beyond Tugboat's Slack notification.
        // These run if `optout` is false.
        $this->arbitraryNotificationOne($result);
        $this->arbitraryTestTwo($result);
        return $result;
```

mmatsoo (approved):
👍

composer.json
markdorison: Will this work with different versions of Guzzle? I ask because of concern over whether there might be other dependencies that require different versions. The wider our requirement can be, the better.

src/Robo/Plugin/Traits/NotifierTrait.php
```diff
@@ -0,0 +1,40 @@
...
+     *   The text string to pass to the Slack API.
+     *
+     * @see https://docs.tugboatqa.com/starter-configs/code-snippets/slack-integration/
+     */
+    public function notifySlack(string $username, string $text): void
+    {
+        // Verify that a Slack webhook URL was provided.
+        $slack_webhook_url = getenv('SLACK_WEBHOOK_URL');
+        if ($slack_webhook_url === false || $slack_webhook_url === '') {
+            $this->yell('Missing Slack Webhook URL from the "SLACK_WEBHOOK_URL" environment variable.');
```
markdorison: @adamzimmermann I created an "Usher" app in Slack and added a webhook that points to #sandbox. Hopefully, if you add it to a repo where you are testing this, it should work.

If it works how we want, we would create a webhook for each repo and point it to the appropriate channel. I will send you the webhook privately.

src/Robo/Plugin/Traits/NotifierTrait.php
```diff
@@ -0,0 +1,40 @@
...
+            return;
+        }
+
+        // Send the Slack webhook call.
+        $client = new Client(['timeout' => 5]);
+        $client->post($slack_webhook_url, [
+            'username' => $username,
+            'text' => $text,
+
```
markdorison:
```suggestion
```

adamzimmermann: Did you miss something here? 🤔

markdorison: Just an empty line that I figured was a stray.

README.md
```diff
@@ -46,6 +46,15 @@ which includes:
...
 1. Generating a login link.
 
+### `robo deploy:drupal`
+
+The `deploy:drupal` command (often available in downstream repos as
+`composer robo deploy:drupal SITENAME DOCROOT`) deploys Drupal in a standardized way
+which includes:
+
+1. `drush deploy`
+1. `drush config:import`
```
apotek, line 56: To me this is a little unclear. drush deploy already include a cache rebuild, an update-db, and a config import. So why the extra callout for config:import? If I'm just being a dummy, please ignore me.

composer.json
apotek: I see your point @markdorison so addressing your concern, would that mean, trying to drop guzzle requirements to the lowest actually required version for our level of php compatibility? Ie, go to version 5 or such? If we are already at the lowest possible supported version, then we are already as wide as we can be.

Guzzle 6 supports php 5.5+ though, so that's pretty wide I think. Maybe it's good where it is.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -6,12 +6,16 @@
 use Robo\Result;
 use Robo\Robo;
 use Robo\Tasks;
+use Usher\Robo\Plugin\Traits\ResultCheckTrait;
```
apotek: I don't see `ResultCheckTrait` being used in this class.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +84,47 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
             // https://github.com/drush-ops/drush/issues/2449#issuecomment-708655673
             ->exec("$appDirPath/vendor/bin/drush config:import --yes")
```
apotek, line 86: Is there any way we can reduce our reliance on hard-coded vendor/bin/drush pathing and either trust that composer did the right things for us (adding vendor/bin to PATH) _or_ if we don't trust composer to do the right thing, at the very least define a constant somewhere so that we aren't redeclaring this path (and our assumptions) all over the place.

src/Robo/Plugin/Commands/DeployCommands.php
apotek: I like @mmatsoo's suggestion.

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +84,47 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
...
+     *
+     * @param \Robo\Result $result
+     *   The result of the task to check.
+     *
+     * @see https://docs.tugboatqa.com/starter-configs/code-snippets/slack-integration/
+     */
+    protected function notifySlackOnFailedBasePreviewBuild(Result $result): void
+    {
+        // Confirm we are in a Tugboat environment.
+        if (getenv('TUGBOAT_PREVIEW_ID') === false) {
```
apotek: I'm surprised `false` is not flagged by phpcs. Shouldn't it be `FALSE`? or does Robo/Usher have different coding standards?

src/Robo/Plugin/Commands/DeployCommands.php
```diff
@@ -73,5 +84,47 @@ public function deployDrupal(string $appDirPath, string $siteName = 'default', s
...
+            return;
+        }
+        // If everything went well there is nothing to do.
+        if ($result->wasSuccessful()) {
+            $this->say('Skipping Slack notification since all tasks completed successfully.');
+            return;
+        }
+        // Build various variables and URLs for the Slack message.
+        $dashboard_url = sprintf(
+            'https://dashboard.tugboatqa.com/%s',
```
apotek: I wish things like domain names weren't strings in operational code. It's not a big deal, but seems like it should be a constant defined in the tooling setup.

src/Robo/Plugin/Traits/NotifierTrait.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Usher\Robo\Plugin\Traits;
+
+use GuzzleHttp\Client;
+use Robo\Result;
+
+/**
+ * Trait to provide notification functionality.
+ */
+trait NotifierTrait
```
apotek: If this trait is actually a slack notifier trait, should we name it SlackNotifierTrait? 

Alternatively, it could be a Notifier interface implemented as a SlackNotifierTrait. But a generic NotifierTrait that is specific to Slack notifications feels off to me.

src/Robo/Plugin/Commands/DeployCommands.php
apotek: Seems like this override could be useful. Why not keep it and make it an option?

apotek (commented):
Mostly questions, because I have a lot to learn. I expect most of my comments / suggestions to be invalid. 

Outside the scope of the PR though I am struck by how much effort we are putting in to dealing with tugboat basically always/mostly being broken.

composer.json
markdorison: Dropping would be good. So would allowing 7.x if our code would work with that!

src/Robo/Plugin/Commands/DeployCommands.php
markdorison: Good catch. I think that was from a previous iteration.

```suggestion
```

src/Robo/Plugin/Commands/DeployCommands.php
markdorison: I am typically in the camp of being as explicit as possible in scripts or code like this so that we aren't ever relying on trusting another system to be configured a certain way ($PATH in this case). I suppose I figure that those characters are free!

src/Robo/Plugin/Commands/DeployCommands.php
markdorison: Usher is using what Robo uses. PSR-12? I will look it up to verify, but we chose not to use Drupal coding standards on Usher since this is not a Drupal project.

src/Robo/Plugin/Commands/DeployCommands.php
apotek: > I am typically in the camp of being as explicit as possible in scripts or code like this so that we aren't ever relying on trusting another system to be configured a certain way ($PATH in this case). I suppose I figure that those characters are free!

🏑 fight! 😆

src/Robo/Plugin/Commands/DeployCommands.php
apotek: > Usher is using what Robo uses. PSR-12? I will look it up to verify, but we chose not to use Drupal coding standards on Usher since this is not a Drupal project.

Got it. I have taken note and will remember.

src/Robo/Plugin/Commands/DeployCommands.php
adamzimmermann: Good idea. They are similar to magic numbers. I'll make this change.

src/Robo/Plugin/Traits/NotifierTrait.php
adamzimmermann: Ah, got it. ✅

README.md
adamzimmermann: Good question. That's a can of worms that is documented in the command itself. 😬 

- https://github.com/ChromaticHQ/usher/blob/2.x/src/Robo/Plugin/Commands/DeployCommands.php#L68

How would you suggest I denote this to avoid future confusion while avoiding duplication?

Add something like `See the code for additional information`?

src/Robo/Plugin/Traits/NotifierTrait.php
markdorison: I am fine with either of those changes!

src/Robo/Plugin/Traits/SlackNotifierTrait.php
```diff
@@ -0,0 +1,102 @@
...
+     * Notify Slack if a base preview build failed.
+     *
+     * @param \Robo\Result $result
+     *   The result of the task to check.
+     * @param bool $force
+     *   If the notification should be forced.
+     *
+     * @see https://docs.tugboatqa.com/starter-configs/code-snippets/slack-integration/
+     */
+    public function notifySlackOnFailedBasePreviewBuild(Result $result, bool $force): void
```
adamzimmermann:
```suggestion
    public function notifySlackOnFailedBasePreviewBuild(Result $result, bool $force = false): void
```

When I make this change, which is what we seemingly would want, I get this error when I run `composer robo`.

`ERROR: Uncaught Symfony\Component\Console\Exception\LogicException: Cannot add a required argument "result" after an optional one "force". in /Users/adamzimmermann/www/usher/vendor/symfony/console/Input/InputDefinition.php:108
`

Any ideas? Clearly Robo is doing some magic. @markdorison I feel like you have encountered this before and shared something with me. I forget what the fix was though.

markdorison: I don't have much to add. It seems like Robo is reordering parameters or injecting their own so we can't safely use optional parameters as they may not end up at the end of the list.

src/Robo/Plugin/Commands/DeployCommands.php
adamzimmermann: I'll see what I can do. Standby!

src/Robo/Plugin/Traits/NotifierTrait.php
adamzimmermann: We had a public method in the trait that Robo was trying to expose as a command. I simply had to make it protected. 🙈

mmatsoo (approved):
I approved this 28 days ago and I'll approve it again!

src/Robo/Plugin/Traits/SlackNotifierTrait.php
```diff
@@ -0,0 +1,102 @@
...
+            return;
+        }
+        // Build various variables and URLs for the Slack message.
+        $dashboard_url = sprintf(
+            '%s/%s',
+            $this->tugboatDashboardUrl,
+            getenv('TUGBOAT_PREVIEW_ID'),
+        );
+        $text = sprintf(
+            "Tugboat Base Preview failed to build for *%s*\n*Preview*: %s\n*Dashboard*: %s",
```
markdorison: @adamzimmermann  What do you think about making these proper markdown links so the notification isn't so verbose with URLs? Made it a list too, but I don't care as much about that.

```suggestion
            "Tugboat Base Preview failed to build for *%s:*\n* [Preview](%s)\n* [Dashboard](%s)",
```

adamzimmermann:
```suggestion
            "Tugboat [Base Preview](%s) failed to build for *%s*",
```
Should we go a step farther? The preview link isn't that valuable, the dashboard gets you where you need to go, and now it's all just one line.

markdorison: > The preview link isn't that valuable

Good point. What about a link to the PR though? Is that a link we can determine?

adamzimmermann: Updated.

markdorison:
@adamzimmermann Can we see the updated message in the sandbox channel?

adamzimmermann:
Screenshot 2022-12-12 at 11 13 53 AM

@markdorison 📷

markdorison (approved):

src/Robo/Plugin/Commands/DeployCommands.php
adamzimmermann: I missed this before, but let's call this a refactor for another day.

adamzimmermann:
I just added some error handling based upon this comment: https://github.com/ChromaticHQ/benz-platform/pull/4349/files#r1046134862

markdorison (approved):

PR #99, opened 2022-12-17
Add Slack Webhook documentation

adamzimmermann:
## Description

We should document how to use these features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,14 @@ which includes:
 1. `drush deploy`
 1. `drush config:import`
 
-#### Options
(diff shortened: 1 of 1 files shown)


mmatsoo (approved):
👍 A lovely sentence, to be sure. Any benefit in adding an example URL or clarifying where to get it, so the developer knows what `SLACK_WEBHOOK_URL` should look like?

adamzimmermann:
@mmatsoo good idea. 💡 

@markdorison handled all of that for me, so perhaps he could contribute a line or two to this PR as well.

markdorison:
Updated!

adamzimmermann:
@markdorison thank you! 🏆

PR #101, opened 2022-12-19
Update CI trigger to point to correct branch

markdorison:
## Description
Update CI trigger to point to correct branch

## Motivation / Context
The CI workflow was running against all PRs as expected but not against the default branch since it has been renamed from `main` -> `2.x`

Review diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:  # yamllint disable-line rule:truthy
   # Trigger the workflow on pushes to the main branch and all pull requests.
   push:
     branches:
-      - main
+      - 2.x
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:  # yamllint disable-line rule:truthy
   # Trigger the workflow on pushes to the main branch and all pull requests.
   push:
     branches:
-      - main
+      - 2.x
```
adamzimmermann:
```suggestion
      - *.x
```
Would this work?

adamzimmermann (approved):

.github/workflows/ci.yml
markdorison: [I think so](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore) ([1](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)). Either that or 
```suggestion
      - **.x
```

adamzimmermann: I think your suggestion is correct. ✅

PR #103, opened 2022-12-27
Removed potentially confusing log message.

markdorison:
## Description
Removed potentially confusing log message.

## Motivation / Context
When viewing Tugboat logs, the message "No Tugboat base preview found." could be misinterpreted when in fact we are just trying to determine if the current preview is a base preview or not, not whether one exists.

I determined that even if we rewrote the message, it probably would not be useful to show up in the logs, so I removed it.

Review PR.

## Screenshots
CleanShot 2022-12-27 at 14 04 21@2x


apotek (approved):
👍

adamzimmermann:
Good call. ✅

PR #107, opened 2023-01-24
Added Drupal status report command.

markdorison:
## Description
Added Drupal status report command.

## Motivation / Context
Allow for easily configurable Drupal status checks in test environments.

TK.

### DIFF
--- a/robo.drupal.example.yml
+++ b/robo.drupal.example.yml
@@ -5,6 +5,8 @@ custom_code_paths:
   - 'web/modules/custom'
   - 'web/themes/custom'
   - 'web/sites/*/themes'
+drupal_status_report_ignore_checks:
+  - update_core
 phpcs_check_extensions:
   - php
   - module
--- a/src/Robo/Plugin/Commands/DrupalStatusReportCommands.php
+++ b/src/Robo/Plugin/Commands/DrupalStatusReportCommands.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Usher\Robo\Plugin\Commands;
+
+use DrupalFinder\DrupalFinder;
+use Robo\Exception\TaskException;
+use Robo\Robo;
+use Robo\Tasks;
+
+/**
+ * Robo commands to work with Drupal's status report.
+ */
+class DrupalStatusReportCommands extends Tasks
+{
+    /**
+     * Drupal root directory.
+     *
+     * @var string
+     */
+    protected $drupalRoot;
+
+    /**
+     * Class constructor.
+     */
+    public function __construct()
+    {
+        // Treat this command like bash -e and exit as soon as there's a failure.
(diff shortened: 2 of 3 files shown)


robo.drupal.example.yml
```diff
@@ -5,6 +5,8 @@ custom_code_paths:
   - 'web/modules/custom'
   - 'web/themes/custom'
   - 'web/sites/*/themes'
+drupal_status_report_ignore_checks:
```
adamzimmermann, line 8: Can we add something to the README that explains what these values should be and how to find them?

src/Robo/Plugin/Commands/DrupalStatusReportCommands.php
```diff
@@ -0,0 +1,65 @@
...
+     * @param int $severity
+     *   The minimum severity level to show. Defaults to 1 (warning).
+     *
+     * @throws \Robo\Exception\TaskException
+     */
+    public function drupalStatusReport($siteDir = 'default', $severity = 1): void
+    {
+        $ignoreList = implode(',', Robo::config()->get('drupal_status_report_ignore_checks'));
+        $result = $this->taskExec("$this->drupalRoot/../vendor/bin/drush status-report --format=json
+            --severity=$severity --ignore=$ignoreList")
```
adamzimmermann, line 50: Didn't realize you could pass all of these flags to the status command. 🔮

src/Robo/Plugin/Commands/DrupalStatusReportCommands.php
```diff
@@ -0,0 +1,65 @@
...
+            ->dir("$this->drupalRoot/sites/$siteDir/")
+            ->printOutput(false)
+            ->run();
+        $drushOutput = trim($result->getOutputData());
+        $reportJson = json_decode($drushOutput);
+        if (!is_array($reportJson) || count($reportJson) > 0) {
+            $this->say($drushOutput);
```
adamzimmermann, line 57: Is there a better method than `say`?

adamzimmermann (approved):
A few nitpicks, but I love that we are adding tools like this.

src/Robo/Plugin/Commands/DrupalStatusReportCommands.php
markdorison: Better in what way?

adamzimmermann: More semantic and we might get fancy color coding? 🤷🏼‍♂️ 

Not a big deal, just wondered if there was another method that would make it have some warning type formatting.

markdorison: Not that I am aware of. This is how we are doing it elsewhere.

PR #110, opened 2023-02-08
Implemented GitHub Status check.

markdorison:
## Description
* Implemented GitHub Status check.
* Refactored Status Report and Config Validation commands to allow them to support multi-sites so that the looping does not need to occur outside of Usher (like in a Bash script.)

## Motivation / Context
* Expose this functionality to all sites using Usher.
* Reduce code duplication.

Tested this branch on a [downstream PR](https://github.com/ChromaticHQ/benz-platform/pull/4560).

## Screenshots
CleanShot 2023-02-09 at 13 25 14@2x

### DIFF
--- a/src/Robo/Plugin/Commands/ValidateConfigCommands.php
+++ b/src/Robo/Plugin/Commands/ValidateConfigCommands.php
@@ -56,6 +56,8 @@ class ValidateConfigCommands extends Tasks
      */
     public function validateDrupalConfig($siteDirs = 'default', array $options = ['set-pr-status' => false]): void
     {
+        $this->io()->title('validate drupal configuration.');
+
         if ($options['set-pr-status']) {
             $this->setGitHubStatusPending(self::GITHUB_STATUS_CHECK_NAME);
         }
--- a/src/Robo/Plugin/Commands/ValidateConfigCommands.php
+++ b/src/Robo/Plugin/Commands/ValidateConfigCommands.php
@@ -5,12 +5,22 @@ namespace Usher\Robo\Plugin\Commands;
 use DrupalFinder\DrupalFinder;
 use Robo\Exception\TaskException;
 use Robo\Tasks;
+use Usher\Robo\Plugin\Traits\GitHubStatusTrait;
 
 /**
  * Robo commands related to changing development modes.
  */
 class ValidateConfigCommands extends Tasks
 {
+    use GitHubStatusTrait;
+
+    /**
+     * The name of the GitHub status check, if set.
+     *
+     * @var string
+     */
+    protected const GITHUB_STATUS_CHECK_NAME = 'ci/configuration-check';
+
     /**
      * Drupal root directory.
      *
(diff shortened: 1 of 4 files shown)


src/Robo/Plugin/Traits/GitHubStatusTrait.php
```diff
@@ -0,0 +1,139 @@
...
+ * Trait to allow setting of status checks on GitHub PRs from Tugboat.
+ */
+trait GitHubStatusTrait
+{
+    /**
+     * The error value when setting a GitHub check status.
+     *
+     * @var string
+     */
+    protected $checkStatusError = 'error';
```
adamzimmermann, line 22: Should any of these be constants? Probably a nitpick?

src/Robo/Plugin/Traits/GitHubStatusTrait.php
```diff
@@ -0,0 +1,139 @@
...
+     *   The descriptive text to set on the check.
+     *
+     * @see https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status
+     */
+    protected function setGitHubStatus(
+        string $state,
+        string $gitHubCheckName,
+        string $checkDescription = ''
+    ): void {
+        $tugboatPreviewID = getenv('TUGBOAT_PREVIEW_ID');
```
adamzimmermann, line 111: Do we need to document these in the README? 

I assume you had to add all of these to the GH Action for Benz?

src/Robo/Plugin/Commands/ValidateConfigCommands.php
```diff
@@ -35,28 +45,45 @@ public function __construct()
     /**
      * Validate Drupal configuration status.
      *
-     * @param string $siteDir
-     *   The Drupal site directory.
-     *
+     * @param string $siteDirs
+     *   A comma separated list of Drupal site directories.
```
adamzimmermann, line 49: Did we do this vs accepting an array for backwards compatibility?

Or did we do this since that makes it easier to pass the value from the CLI?

adamzimmermann (approved):

src/Robo/Plugin/Traits/GitHubStatusTrait.php
markdorison: No, this runs _in_ Tugboat, so these are all available.

src/Robo/Plugin/Commands/ValidateConfigCommands.php
markdorison: CLI. This is a publicly-exposed command usually called from the command-line, so I wanted to keep that possible.

src/Robo/Plugin/Traits/GitHubStatusTrait.php
adamzimmermann: Oh right. 🙈

src/Robo/Plugin/Commands/ValidateConfigCommands.php
adamzimmermann: I figured that was the idea half-way through my comment.

src/Robo/Plugin/Traits/GitHubStatusTrait.php
markdorison: [Constants are only available on traits in PHP 8.2](https://php.watch/versions/8.2/constants-in-traits). When we can drop support for < 8.2 then we could swap for constants!

PR #111, opened 2023-02-09
Drop PHP 7.4 and 8.0 support.

markdorison:
## Description
* Drop PHP 7.4 and 8.0 support.
* Updated some code with PHP 8.0/8.1 goodies.
* This will likely be the first commit on the `3.x` branch.

## Motivation / Context
[PHP 7.4 is now out of the security support window](https://www.php.net/supported-versions.php) and 8.0 is approaching it. Let's keep things moving with a new major version.

Tests should pass.


src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -131,12 +125,9 @@ protected function jobRunCodingStandards(bool $applyFixes = false): Result
...
     /**
      * Get the list of custom code paths to check.
-     *
-     * @return string
-     *   A space separated list of custom code paths.
      */
     protected function getCustomCodePaths(): string
     {
-        return implode(' ', $this->getRequiredRoboConfigArrayFor('custom_code_paths'));
+        return implode(separator: ' ', array: $this->getRequiredRoboConfigArrayFor(key: 'custom_code_paths'));
```
adamzimmermann:
```suggestion
        return implode(
          separator: ' ',
          array: $this->getRequiredRoboConfigArrayFor(key: 'custom_code_paths'),
        );
```
?

src/Robo/Plugin/Traits/RoboConfigTrait.php
```diff
@@ -15,14 +16,15 @@ trait RoboConfigTrait
...
      */
     protected function getRequiredRoboConfigArrayFor(string $key): array
     {
-        $configValue = $this->getRequiredRoboConfigValueFor($key);
-        $this->validateRoboConfigValueMatchesType($configValue, 'array', $key);
+        $configValue = $this->getRequiredRoboConfigValueFor(key: $key);
+        $this->validateRoboConfigValueMatchesType(
+            configValue: $configValue,
+            expectedType: ConfigTypes::array,
+            key: $key
```
adamzimmermann:
```suggestion
            key: $key,
```

adamzimmermann: I see this below too, perhaps what I'm suggesting is wrong?

src/Robo/Plugin/Traits/RoboConfigTrait.php
```diff
@@ -81,25 +82,18 @@ private function getRequiredRoboConfigValueFor(string $key)
...
-     *
-     * @return bool
-     *   TRUE if configuration value matches the expected type.
-     *
      * @throws \Robo\Exception\TaskException
      */
-    private function validateRoboConfigValueMatchesType($configValue, string $expectedType, string $key): bool
-    {
+    private function validateRoboConfigValueMatchesType(
+        mixed $configValue,
```
adamzimmermann, line 88: TIL that `mixed` is valid.

src/Robo/Plugin/Traits/SlackNotifierTrait.php
```diff
@@ -79,16 +74,13 @@ protected function sendSlackNotification(string $username, string $text): void
         try {
             $client = new Client(['timeout' => 5]);
             $client->post($slack_webhook_url, ['body' => json_encode($payload, JSON_THROW_ON_ERROR)]);
-        } catch (RequestException $exception) {
+        } catch (RequestException) {
```
adamzimmermann, line 77: A question for here and other places, should we be doing anything with the exception variable? Logging it could be helpful? I realize that is outside the scope of this PR though.

src/Robo/Plugin/Traits/SlackNotifierTrait.php
```diff
@@ -79,16 +74,13 @@ protected function sendSlackNotification(string $username, string $text): void
...
             $this->yell('Slack webhook request failed.');
         }
     }
 
     /**
      * Determine if we are in a Tugboat base preview.
-     *
-     * @return bool
```
adamzimmermann, line 90: So cool that we don't have to add this, we can let the code speak for itself. 💯

adamzimmermann (approved):
A few questions, but overall this looks great. Excited to see us moving this forward. 👏🏼

src/Robo/Plugin/Traits/RoboConfigTrait.php
markdorison: [New in 8.0](https://stitcher.io/blog/new-in-php-8#new-mixed-type-rfc)! We could also maybe use a union type here like `string|array`. 🤔

src/Robo/Plugin/Traits/SlackNotifierTrait.php
markdorison: If there is data in there that is useful, then yes, we should. Since this is a CLI command and I am outputting an error, I didn't see a need to use any of the data in the exception.

src/Robo/Plugin/Traits/RoboConfigTrait.php
markdorison: I guess it would need to be something like `string|int|float|bool|null|array`. `mixed` is probably best.
CleanShot 2023-05-31 at 15 51 59@2x

adamzimmermann: 👍🏼

markdorison:
@adamzimmermann Thanks for the feedback and suggestions! Pushed up some changes.

src/Robo/Plugin/Traits/RoboConfigTrait.php
adamzimmermann: I'm guessing that the trailing comma for named arguments is not valid based upon you resolving this?

markdorison: I committed it! Let me fix up the other instances.

adamzimmermann: Ah, I misread what GitHub was telling me. My bad!

PR #114, opened 2023-02-21
Add cache:clear bin config before config:status check.

markdorison:
## Description
Add `cache:clear bin config` before `config:status` check

## Motivation / Context
Attempting to solve edge cases where config:status is showing a diff where we do not believe one exists.

TK.

### DIFF
--- a/src/Robo/Plugin/Commands/ValidateConfigCommands.php
+++ b/src/Robo/Plugin/Commands/ValidateConfigCommands.php
@@ -63,6 +63,8 @@ class ValidateConfigCommands extends Tasks
         }
         $sites = explode(',', $siteDirs);
         foreach ($sites as $siteDir) {
+            // Clear the "config" cache bin before we verify config status to
+            // improve the accuracy of this check.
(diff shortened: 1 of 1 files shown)


adamzimmermann (approved):
I love the comments and links 💯 , but should we also explain how we got here a bit and why this was added?

mmatsoo (approved):
👍

adamzimmermann:
@markdorison could we:

- Run `drush config:status` and see a mis-match reported on a failing PR
- Run `drush config:get` and see what it reports.
- Look at config files to see what values they have.
- Look in config cache in the database to see what it reports.

See where things break down and the diff can be found. Then report this as a bug somewhere or restart one of the existing conversations I found? At the very least we could write a blog post about it, as this feels like thing thing that needs some attention shed on it.

markdorison:
@adamzimmermann You are suggesting that is what the command should do?

adamzimmermann:
@markdorison no, I'm just hoping to be able to better answer/understand why we need this change with some facts to help our own understanding and as something to share with the community and possibly provide as a fix somewhere.

PR #116, opened 2023-03-14
Add command for refreshing all configured sites.

markdorison:
## Description
Add command for refreshing all configured sites.

## Motivation / Context
Discussion with @alexandar87.

Tested locally.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -207,15 +207,19 @@ class DevelopmentModeBaseCommands extends Tasks
      *
      * @param string $siteName
      *   The Drupal site name.
+     * @param bool $skipLandoStart
+     *   If TRUE, skip starting Lando.
      *
      * @return \Robo\Result
      *   The result of the set of tasks.
      */
-    protected function devRefreshDrupal(string $siteName = 'default'): Result
+    protected function devRefreshDrupal(string $siteName = 'default', bool $skipLandoStart = false): Result
     {
         $this->io()->title('development environment refresh. 🦄✨');
         $result = $this->taskComposerInstall()->run();
-        $result = $this->taskExec('lando')->arg('start')->run();
+        if (!$skipLandoStart) {
+            $result = $this->taskExec('lando')->arg('start')->run();
+        }
         // There isn't a great way to call a command in one class from another.
         // https://github.com/consolidation/Robo/issues/743
         // For now, it seems like calling robo from within robo works.
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -24,15 +24,48 @@ class DevelopmentModeCommands extends DevelopmentModeBaseCommands
      *
      * @param string $siteName
      *   The Drupal site name.
+     * @option skip-lando-start
+     *   Skip starting Lando.
      *
      * @aliases magic
      *
      * @return \Robo\Result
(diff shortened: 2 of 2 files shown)


src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -207,15 +207,19 @@ public function frontendDevDisable($siteDir = 'default', array $opts = ['yes|y'
      *
      * @param string $siteName
      *   The Drupal site name.
+     * @param bool $skipLandoStart
```
adamzimmermann, line 210: Variables named in the affirmative are much easier to understand IMO.

```suggestion
     * @param bool $startLando
```
What about this?

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -24,15 +24,48 @@ class DevelopmentModeCommands extends DevelopmentModeBaseCommands
...
+     * @return \Robo\Result
+     *   The result of the set of tasks.
+     */
+    public function devRefreshAll(array $options = ['skip-sites' => '', 'skip-lando-start' => false]): Result
+    {
+        $skipLandoStart = $options['skip-lando-start'];
+        $siteNames = $this->getAllSiteNames();
+        $result = null;
+        foreach ($siteNames as $siteName) {
+            if (in_array($siteName, explode(',', $options['skip-sites']), true)) {
```
adamzimmermann, line 63: Do we need/want the strict checking?

(I didn't even know that was an option and had to google it to understand what it even did.)

adamzimmermann (approved):
Looking good. A few nitpicks and questions.

Is it worth adding something explaining how this might take a while if many sites are involved? I'm on the fence about that, as that is a consuming repo implementation detail to some extent.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: > (I didn't even know that was an option and had to google it to understand what it even did.)

@adamzimmermann Did you find something because this was just my creation to be able to [skip a site on Benz](https://github.com/ChromaticHQ/benz-platform/blob/e169c6736974f33b2ea3fcece7192de34051a718/.sites.config.yml#L61-L66). 😆

adamzimmermann: I was referring to the strict checking option that the third argument provides.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: I hear you, but since I [implemented this as a flag](https://github.com/ChromaticHQ/usher/pull/116/files#diff-f97502affc46a87c9b2819746c440a2b3bd57e1738085378386a5da44066cf28R35), it makes sense to have `false` be the default and only include the flag when you want to change the behavior. Since starting Lando is currently the default, I didn't want to change that.

We could do a bigger refactor to change all this, but I don't have time to address that at the moment.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: Ohhhh. Hah! I learned about it myself doing this. New warnings teaching us things!

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
adamzimmermann: Ah, makes sense. 👍🏼

PR #121, opened 2023-04-20
Fix cache:rebuild running incorrectly on D7 sites.

markdorison:
## Description
Fix cache:rebuild running incorrectly on D7 sites.

## Motivation / Context
Error on D7 site.

Tested manually.

cc/ @mmatsoo

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -10,6 +10,7 @@ use Robo\Tasks;
 use Symfony\Component\Yaml\Exception\ParseException;
 use Symfony\Component\Yaml\Yaml;
 use Usher\Robo\Plugin\Traits\DatabaseDownloadTrait;
+use Usher\Robo\Plugin\Traits\DrupalVersionTrait;
 use Usher\Robo\Plugin\Traits\SitesConfigTrait;
 
 /**
@@ -21,6 +22,7 @@ class DevelopmentModeBaseCommands extends Tasks
 {
     use DatabaseDownloadTrait;
     use SitesConfigTrait;
+    use DrupalVersionTrait;
 
     /**
      * Drupal root directory.
@@ -124,10 +126,12 @@ class DevelopmentModeBaseCommands extends Tasks
                 ->run();
             $result = $this->taskExec('rm')->args($dbPath)->run();
 
-            $result = $this->taskExec("$this->vendorDirectory/bin/drush")
-                ->arg('cache:rebuild')
-                ->dir("$this->drupalRoot/sites/$siteName")
-                ->run();
+            if (!$this->drupalVersionIsD7($this->drupalRoot)) {
+                $result = $this->taskExec("$this->vendorDirectory/bin/drush")
+                    ->arg('cache:rebuild')
+                    ->dir("$this->drupalRoot/sites/$siteName")
+                    ->run();
+            }
         }
         return $result;
     }
--- a/src/Robo/Plugin/Traits/DrupalVersionTrait.php
+++ b/src/Robo/Plugin/Traits/DrupalVersionTrait.php
@@ -19,7 +19,7 @@ trait DrupalVersionTrait
         $bootstrapPath = "$drupalRootPath/includes/bootstrap.inc";
         if (file_exists($bootstrapPath)) {
             include_once($bootstrapPath);
-            if (substr(VERSION, 0, 2) === '7.') {
+            if (defined('VERSION') && substr(VERSION, 0, 2) === '7.') {
                 return true;
             }
         }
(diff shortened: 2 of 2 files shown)


mmatsoo (approved):
👍 I have not tested against a D7 site, but having gone through yesterday's Intuit challenges, the changes here look like they overcome the cache-clearing difficulties I discovered.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -124,10 +126,12 @@ public function databaseRefreshTugboat(): Result
...
-            $result = $this->taskExec("$this->vendorDirectory/bin/drush")
-                ->arg('cache:rebuild')
-                ->dir("$this->drupalRoot/sites/$siteName")
-                ->run();
+            if (!$this->drupalVersionIsD7($this->drupalRoot)) {
+                $result = $this->taskExec("$this->vendorDirectory/bin/drush")
+                    ->arg('cache:rebuild')
+                    ->dir("$this->drupalRoot/sites/$siteName")
+                    ->run();
+            }
```
adamzimmermann, line 134: Do we need an `else` for D7?

src/Robo/Plugin/Traits/DrupalVersionTrait.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+trait DrupalVersionTrait
+{
+    /**
+     * Check if version of Drupal in question is Drupal 7.
+     *
+     * @param string $drupalRootPath
+     *   The path to the Drupal root.
+     */
+    protected function drupalVersionIsD7(string $drupalRootPath): bool
```
adamzimmermann, line 16: Could we do something like this?

- https://www.liquidweb.com/kb/version-drupal-running/

Seems slightly less brittle?

I don't want us to over think this though either if this works. 🤷🏼‍♂️

adamzimmermann (approved):
A few questions, but looks like it works as is. ✅

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: This was [added (relatively) recently](https://github.com/ChromaticHQ/usher/commit/4d02216282412f8133de6accee4152c87ffe9b5b) to solve a problem on D8/9/10. I didn't want to add a step for D7 if I didn't know that it was needed.

adamzimmermann: Ah, this is that cache clear. 👍🏼

src/Robo/Plugin/Traits/DrupalVersionTrait.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        // This line has not changed since 2013, so it seems safe to base our version check off of.
+        // https://git.drupalcode.org/project/drupal/-/blame/7.x/COPYRIGHT.txt
+        $copyrightString = 'All Drupal code is Copyright 2001 - 2013 by the original authors.';
+        $copyrightPath = "$drupalRootPath/COPYRIGHT.txt";
+        if (file_exists($copyrightPath) && $copyrightContents = file_get_contents($copyrightPath)) {
+            if (str_contains($copyrightContents, $copyrightString)) {
```
markdorison:
```suggestion
         // @see https://git.drupalcode.org/project/drupal/-/blob/7.x/includes/bootstrap.inc
        $bootstrapPath = "$drupalRootPath/includes/bootstrap.inc";
        if (file_exists($bootstrapPath)) {
            include_once($bootstrapPath);
            if (substr(VERSION, 0, 2) === '7.') {
```

@adamzimmermann Is this less brittle? Relying on the version number definitely *seems* more reliable on the surface, but the version constant changes value frequently vs the string that hasn't changed in 10 years. What do you think?

adamzimmermann: > but the version constant changes value frequently

Your check accounts for the change though, right?

Or is there another element of risk that I'm not considering?

markdorison: It _should_... I am probably not being trusting enough.

adamzimmermann: I always support being paranoid with code.

markdorison: Alright, I am going for it. 😆

src/Robo/Plugin/Traits/DrupalVersionTrait.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param string $drupalRootPath
+     *   The path to the Drupal root.
+     */
+    protected function drupalVersionIsD7(string $drupalRootPath): bool
+    {
+        // @see https://git.drupalcode.org/project/drupal/-/blob/7.x/includes/bootstrap.inc
+        $bootstrapPath = "$drupalRootPath/includes/bootstrap.inc";
+        if (file_exists($bootstrapPath)) {
+            include_once($bootstrapPath);
+            if (substr(VERSION, 0, 2) === '7.') {
```
markdorison: @adamzimmermann Ahh. PHPStan really doesn't like the use of this constant. I suppose I understand.

CleanShot 2023-04-21 at 10 39 47@2x

PR #122, opened 2023-05-01
Add check for presence of php_current_version.

markdorison:
## Description
Add check for the presence of php_current_version.

## Motivation / Context
Provide better feedback if the command is not adequately configured.

Tested locally.

## Screenshots
CleanShot 2023-05-01 at 14 24 35@2x

### DIFF
--- a/src/Robo/Plugin/Commands/ToolingCommands.php
+++ b/src/Robo/Plugin/Commands/ToolingCommands.php
@@ -119,7 +119,7 @@ class ToolingCommands extends Tasks
     protected function getConfig(string $key)
     {
         $configValue = Robo::config()->get($key);
-        if (empty($configValue)) {
+        if (!isset($configValue)) {
             throw new TaskException($this, "Key $key not found in Robo config file robo.yml.");
         }
         return $configValue;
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/ToolingCommands.php
```diff
@@ -106,4 +106,22 @@ protected function updateRoboConfig(string $key, string $value): void
...
+     * @param string $key
+     *   The key of the configuration to load.
+     *
+     * @return mixed
+     *   A configuration value.
+     */
+    protected function getConfig(string $key)
+    {
+        $configValue = Robo::config()->get($key);
+        if (!isset($configValue)) {
```
adamzimmermann, line 122: Wouldn't it always be set, but it could be empty?

Or does `get()` have a return type that makes this check work?

adamzimmermann (approved):
Approving pending my question.

I like the extra check and no failing silently. 👏🏼

src/Robo/Plugin/Commands/ToolingCommands.php
markdorison: I just tested and confirmed that if the key is present, but not set, it returns NULL and therefore fails this check as expected. That said, I am going to switch this to use `empty()` to be a bit more defensive, just in case..

markdorison: Just kidding. PHPStan _hates_ empty. Switched back to isset() since it worked in my testing.

PR #123, opened 2023-05-05
Created RoboConfigTrait to abstract logic related to retrieving robo.yml config

markdorison:
## Description
* Created `RoboConfigTrait` to abstract logic related to retrieving and validating configuration stored in `robo.yml`.
* Updated commands to use this new trait.

## Motivation / Context
This began with a small code smell when I didn't like that [`getConfig()` in `ToolingCommand`](https://github.com/ChromaticHQ/usher/blob/7a6b516eedf5aa186097ce025e84f8e2fd49622e/src/Robo/Plugin/Commands/ToolingCommands.php#L119-L126) was duplicating code used elsewhere and it snowballed from there.

https://github.com/ChromaticHQ/usher/blob/7a6b516eedf5aa186097ce025e84f8e2fd49622e/src/Robo/Plugin/Commands/ToolingCommands.php#L119-L126

Tests should pass.

### DIFF
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -23,61 +23,12 @@ class CICommands extends Tasks
     protected const PHPCS_DEFAULT_PHP_VERSION = '8.1';
 
     /**
-     * A comma separated list of the file extensions PHPCS should check.
-     *
-     * @var string
-     */
-    protected $phpcsCheckExtensions;
-
-    /**
-     * A comma separated list of the paths PHPCS should ignore.
-     *
-     * @var string
-     */
-    protected $phpcsIgnorePaths;
-
-    /**
-     * A space separated list of custom code paths.
-     *
-     * @var string
-     */
-    protected $customCodePaths;
-
-    /**
-     * A comma separated list of PHPCS standards.
-     *
-     * @var string
-     */
-    protected $phpcsStandards;
-
-    /**
-     * The PHP version to lint against for support.
-     *
-     * @var string
-     */
-    protected $phpcsPhpVersion;
-
-    /**
-     * Boolean indicating whether Twig files should be linted.
-     *
-     * @var bool
-     */
-    protected $lintTwigFiles;
-
-    /**
      * RoboFile constructor.
      */
     public function __construct()
     {
         // Treat this command like bash -e and exit as soon as there's a failure.
         $this->stopOnFail();
-
-        $this->phpcsCheckExtensions = implode(',', Robo::config()->get('phpcs_check_extensions'));
-        $this->phpcsIgnorePaths = implode(',', $this->getRoboConfigArrayFor('phpcs_ignore_paths'));
-        $this->customCodePaths = implode(' ', $this->getRoboConfigArrayFor('custom_code_paths'));
-        $this->phpcsStandards = implode(',', $this->getRoboConfigArrayFor('phpcs_standards'));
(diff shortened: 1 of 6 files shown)


src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -145,41 +100,43 @@ public function jobFixCodingStandards(): Result
      */
     protected function jobRunCodingStandards(bool $applyFixes = false): Result
     {
+        $customCodePaths = $this->getCustomCodePaths();
+        $phpcsStandards = implode(',', $this->getRequiredRoboConfigArrayFor('phpcs_standards'));
+        $phpcsCheckExtensions = implode(',', $this->getRequiredRoboConfigArrayFor('phpcs_check_extensions'));
+        $phpcsIgnorePaths = implode(',', $this->getRequiredRoboConfigArrayFor('phpcs_ignore_paths'));
+        $phpcsPhpVersion = Robo::config()->get('phpcs_php_version', $this::PHPCS_DEFAULT_PHP_VERSION);
```
adamzimmermann, line 107: I kind of like some of these things that require "processing" i.e. calling `implode` or things that are "global" being set in the constructor to allow re-use of them, but perhaps that is over-optimizing?

src/Robo/Plugin/Traits/RoboConfigTrait.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param string $key
+     *   The key of the configuration to load.
+     *
+     * @return string
+     *   A configuration string.
+     */
+    protected function getRequiredRoboConfigStringFor(string $key): string
+    {
+        $configValue = $this->getRequiredRoboConfigValueFor($key);
+        $this->validateRoboConfigValueMatchesType($configValue, 'string', $key);
```
adamzimmermann, line 41: Are there constants for the types? e.g. `string`?

src/Robo/Plugin/Traits/RoboConfigTrait.php
```diff
@@ -0,0 +1,107 @@
...
+     *   The type we are expecting the value to be of.
+     * @param string $key
+     *   The key of the configuration.
+     *
+     * @return bool
+     *   TRUE if configuration value matches the expected type.
+     *
+     * @throws \Robo\Exception\TaskException
+     */
+    private function validateRoboConfigValueMatchesType($configValue, string $expectedType, string $key): bool
```
adamzimmermann, line 96: I love it!

adamzimmermann (approved):
A few questions, but also looks good as is too.

Approving and we can hash out any adjustments in comments.

src/Robo/Plugin/Commands/CICommands.php
markdorison: The issue I ran into is that the constructor is called even when just calling `robo` without a command so that it can discover and list all the commands. This works fine if every single one of the configuration values is set or optional, but if you want to configure one as required and throw an exception if it is not set, you would be throwing an exception in a scenario when the actual command is not even being run.

For that reason, I decided to ditch this approach of configuring things in the constructor _especially_ in classes that expose multiple commands.

src/Robo/Plugin/Traits/RoboConfigTrait.php
markdorison: Not that I know of, but when we require PHP 8.1 then we could use an enum for this.

src/Robo/Plugin/Commands/CICommands.php
adamzimmermann: Ah, that makes sense. ✅

PR #130, opened 2023-05-12
Update handling of $options arrays to use list() for variable assignment

markdorison:
## Description
Update handling of `$options` arrays to use `list()` for variable assignment

## Motivation / Context
Less instances calling into `$options` arrays using key strings.

Tests should pass.


adamzimmermann (approved):
👏🏼

PR #142, opened 2023-06-08
Added support for DDEV.

markdorison:
## Description
* Added support for DDEV.

## Motivation / Context
Allow projects that use dev:refresh commands to use either DDEV or Lando.

Test on a project pointing to this branch.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -172,8 +172,6 @@ class DevelopmentModeBaseCommands extends Tasks
      *   Specify local development enviroment: ddev, lando.
      * @param string $siteDir
      *   The Drupal site directory name.
-     *
-     * @option lando Whether to run the automatic fixer or not.
      */
     public function drupalLoginLink(
         string $environmentType,
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -9,6 +9,7 @@ use Robo\Robo;
 use Robo\Tasks;
 use Symfony\Component\Yaml\Exception\ParseException;
 use Symfony\Component\Yaml\Yaml;
+use Usher\Robo\Plugin\Enums\LocalDevEnvironmentTypes;
 use Usher\Robo\Plugin\Traits\DatabaseDownloadTrait;
 use Usher\Robo\Plugin\Traits\DrupalVersionTrait;
 use Usher\Robo\Plugin\Traits\SitesConfigTrait;
@@ -62,6 +63,34 @@ class DevelopmentModeBaseCommands extends Tasks
     }
 
     /**
+     * Refresh a site database in DDEV.
+     *
+     * @param string $siteName
+     *   The Drupal site name.
+     */
+    public function databaseRefreshDdev(string $siteName = 'default'): Result
+    {
+        $this->io()->title('DDEV database refresh.');
+
+        $dbPath = $this->databaseDownload($siteName);
+
+        $this->io()->section("importing $siteName database.");
+        $this->say("Importing $dbPath");
+        $this->taskExec('ddev')
+            ->arg('import-db')
+            ->option('target-db', $siteName == 'default' ? 'db' : $siteName)
+            ->option('src', $dbPath)
+            ->run();
+
+        $this->say("Deleting $dbPath");
+        $this->taskExec('rm')->args($dbPath)->run();
+        return $this->drushDeployWith(
+            localEnvironmentType: LocalDevEnvironmentTypes::DDEV,
+            siteDir: $siteName,
+        );
+    }
+
+    /**
      * Refresh a site database in Lando.
      *
      * @param string $siteName
@@ -85,7 +114,10 @@ class DevelopmentModeBaseCommands extends Tasks
 
         $this->say("Deleting $dbPath");
         $this->taskExec('rm')->args($dbPath)->run();
-        return $this->drushDeployLando($siteName);
+        return $this->drushDeployWith(
+            localEnvironmentType: LocalDevEnvironmentTypes::LANDO,
+            siteDir: $siteName,
+        );
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -197,103 +223,45 @@ public function frontendDevDisable(string $siteDir = 'default', array $opts = ['
...
+    protected function devRefreshDrupal(
+        LocalDevEnvironmentTypes $environmentType,
+        string $siteName = 'default',
+        bool $startLocalEnv = false,
+    ): Result {
         $this->io()->title('development environment refresh. 🦄✨');
         $result = $this->taskComposerInstall()->run();
-        if (!$skipLandoStart) {
-            $result = $this->taskExec('lando')->arg('start')->run();
+        if ($startLocalEnv) {
```
adamzimmermann, line 232: Should everything in this `if` statement be its own method?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -197,103 +223,45 @@ public function frontendDevDisable(string $siteDir = 'default', array $opts = ['
...
-                $landoCfg = Yaml::parseFile($landoConfigPath);
-            } catch (ParseException) {
-                $this->say("Unable to load Lando config from $landoConfigPath.");
-            }
-            // Break out of the loop if valid configuration was found.
-            if ($landoCfg !== false) {
-                $this->say("Found Lando config at $landoConfigPath.");
-                break;
-            }
+        if ($environmentType == LocalDevEnvironmentTypes::LANDO) {
```
adamzimmermann, line 245: Should we have a generic `databaseRefresh` command that has this logic inside it?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -303,14 +271,14 @@ protected function drushDeployLando(string $siteDir = 'default'): Result
...
-            ->exec("lando drush deploy --yes")
+            ->exec("$localEnvironmentType->value drush @$siteDir.$localEnvironmentType->value deploy --yes")
             // Import the latest configuration again. This includes the latest
             // configuration_split configuration. Importing this twice ensures that
             // the latter command enables and disables modules based upon the most up
             // to date configuration. Additional information and discussion can be
             // found here:
             // https://github.com/drush-ops/drush/issues/2449#issuecomment-708655673
-            ->exec("lando drush config:import --yes")
+            ->exec("$localEnvironmentType->value drush @$siteDir.$localEnvironmentType->value config:import --yes")
```
adamzimmermann, line 279: How does this work if the local environment runs `drush` natively`? Does the ` ` space before `drush` cause an issue?

src/Robo/Plugin/Enums/LocalDevEnvironmentTypes.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Usher\Robo\Plugin\Enums;
+
+enum LocalDevEnvironmentTypes: string
```
adamzimmermann, line 5: Big picture idea. Should each of these environment types be classes that extend a base class? Then they could override methods as needed too, and encapsulate all of their custom needs?

adamzimmermann (commented):
Some questions and comments, but overall this looks like a win. I'm a touch worried about the complexity and long-term maintainability of some of this though. Those refactors could be a follow-up PR if we need this as is though.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: Which logic? I just want to make sure I am answering the right question. Are you suggesting combining `databaseRefreshLando()` and `databaseRefreshDdev()`?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: Are you suggesting that from a readability perspective or something else? I didn't go that far in this PR since it was just a line of code each.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
adamzimmermann: > Are you suggesting combining databaseRefreshLando() and databaseRefreshDdev()?

Yes. However, my comment at the end of this PR might change the direction? 🤔

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
adamzimmermann: Single responsibility principle I suppose. 

If you don't want to go that far in this PR that's fine though.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: This seems like the opposite approach of your suggestion above RE: breaking things out to separate methods. Am I misunderstanding?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: > How does this work if the local environment runs drush natively?

* This helper method is only called by `databaseRefreshLando()` and `databaseRefreshDdev()`.
* It has a required argument of `LocalDevEnvironmentTypes` so it would have to be called with a valid option.
* If we added support for other `LocalDevEnvironmentTypes` (like local/native) then we would need to update the logic at that time.

src/Robo/Plugin/Enums/LocalDevEnvironmentTypes.php
markdorison: I see where you are going and I like it if I thought we would keep supporting both long-term but I can also imagine us dropping Lando support if we move everything over. I don't want to over-architect things at this stage until we figure out the direction we are moving.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
adamzimmermann: You are correct. They are incongruent approaches. I'm not sure which is best off-hand, I was just trying to think of ways to make it better as I went through the PR.

src/Robo/Plugin/Enums/LocalDevEnvironmentTypes.php
adamzimmermann: 👍🏼 Makes sense.

adamzimmermann (approved):
Approving and letting you decide what else if anything you want to do.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: All good! You had me 😵

goranmiric (approved):

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: FWIW, I started working on this and I ended up feeling like I just traded one mess for another. [Leaving this patch here for posterity](https://gist.github.com/markdorison/1f477c1edb22a108a2b165c89b67f050) but I abandoned this approach for now.

PR #146, opened 2023-06-22
Update Rector config.

markdorison:
## Description
* Update rector config.
* Ignore troublesome rule temporarily.
* Apply fixes.

## Motivation / Context
Kepp all tests passing.

Tests pass.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,9 +30,9 @@
         }
     },
     "scripts": {
-        "php-check": "composer robo job:check-coding-standards",
-        "php-fix": "composer robo job:fix-coding-standards",
-        "php-stan": "composer robo job:run-static-analysis",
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -30,15 +30,15 @@
         }
     },
     "scripts": {
-        "php-check": "composer robo job:check-coding-standards",
-        "php-fix": "composer robo job:fix-coding-standards",
-        "php-stan": "composer robo job:run-static-analysis",
+        "phpcs": "composer robo job:check-coding-standards",
```
adamzimmermann, line 33: Does this affect any repos that use Usher? I think the answer is no, correct?

adamzimmermann (approved):
👏🏼

composer.json
markdorison: No, it should not.

PR #148, opened 2023-06-29
Update phpcsversion to use php_current_version.

markdorison:
## Description
Deprecate `phpcs_php_version` and rely on `php_current_version` for PHPCS as well.

## Motivation / Context
Simplification. Lowers the chance of confusion.


adamzimmermann (approved):
💯

PR #164, opened 2023-09-15
Resolve mariadb for mysql contextually

### LINKED ISSUE #159: Make databaseRefreshTugboat() able to handle non-mysql database client names
## Description

Given that tugboatQA now has deprecated the use of the `mysql` command in their images in favor of `mariadb`, the following lines of code need to be able to handle this change:

```php
            $result = $this->taskExec('mysql')
                ->option('-h', 'mariadb')
                ->option('-u', 'tugboat')
                ->option('-ptugboat')
                ->option('-e', "drop database if exists $dbName; create database $dbName;")
                ->run();
```

The wrapping code does some auto-discovery of credentials, and it would be ideal if we could also auto-discover from the Drupal settings whether to use `mysql` or `mariadb` (or any other db). 

However, Drupal appears not to support a `mariadb://` connection string, so we are stuck with either:

1) Running an exec against the system like `which mysql` to see if mysql is installed and then flip to mariadb if not (or vice a versa)
2) Setting the db client command in robo.yml.

## Issue Owner
@apotek /  [JFK Library](https://jfklfoundation.atlassian.net/browse/CJFK-513)

## Acceptance Criteria

- [ ] All users of Usher can easily use `mariadb` instead of `mysql` when running the `databaseRefreshTugboat()` command (or, ideally, any refresh db command). 

## Proposed Solution

1. Discover which command to do based on binaries available on the current environment, or
2. Configure which command to use in robo.yml.

## Alternatives Considered

Auto-discover from the Drupal db connection string.

apotek:
## Description

Adds a Robo task that provides a facility for being able to choose which binary to use depending on environment.

## Motivation / Context

I wanted to get ahead of the fact that tugboat will soon be switching to mariadb binaries with no symlinking from mysql. 

Which made me think it would be good to create a Robo task/utility for being able to switch between binaries, depending on the operating environment. 

I used this also an a chance to get to know Robo just a bit. I could obviously have implemented this in a simpler way just for our use case.

I was going to contribute the "Alternatives" task as a Robo task to packagist and then import it, but I felt at this early stage it might be better to keep things simple and just add it to Usher. 

## Sounds good, but why are there so many changes to `databaseRefreshTugboat`?
I noticed that there was no consistent handling of "result" in that function. Result kept getting overwritten, meaning that the returned result at the end was basically meaningless. On top of that, if an exception occurred in attempting to get the dbPath, and there was only one site, there would be a return of NULL, which broke the function signature. 

## Lando?
I also added a .lando file so I could easily test some robo stuff _in situ_ without having to import usher into another project just to test something. 

Resolves: #159

### DIFF
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -1,0 +1,262 @@
+name: usher
+type: php
+docroot: ""
+php_version: "8.2"
+webserver_type: nginx-fpm
+xdebug_enabled: false
+additional_hostnames: []
+additional_fqdns: []
+use_dns_when_possible: true
+composer_version: "2"
+web_environment: []
+nodejs_version: "18"
+omit_containers:
+    - db
+
+# Key features of ddev's config.yaml:
+
+# name: <projectname> # Name of the project, automatically provides
+#   http://projectname.ddev.site and https://projectname.ddev.site
+
+# type: <projecttype>  # drupal6/7/8, backdrop, typo3, wordpress, php
+
+# docroot: <relative_path> # Relative path to the directory containing index.php.
+
+# php_version: "8.1"  # PHP version to use, "5.6", "7.0", "7.1", "7.2", "7.3", "7.4", "8.0", "8.1", "8.2"
+
+# You can explicitly specify the webimage but this
+# is not recommended, as the images are often closely tied to ddev's' behavior,
+# so this can break upgrades.
+
+# webimage: <docker_image>  # nginx/php docker image.
+
+# database:
+#   type: <dbtype> # mysql, mariadb, postgres
+#   version: <version> # database version, like "10.4" or "8.0"
+#   mariadb versions can be 5.5-10.8 and 10.11, mysql versions can be 5.5-8.0
+#   postgres versions can be 9-15.
+
+# router_http_port: <port>  # Port to be used for http (defaults to global configuration, usually 80)
+# router_https_port: <port> # Port for https (defaults to global configuration, usually 443)
+
+# xdebug_enabled: false  # Set to true to enable xdebug and "ddev start" or "ddev restart"
+# Note that for most people the commands
+# "ddev xdebug" to enable xdebug and "ddev xdebug off" to disable it work better,
+# as leaving xdebug enabled all the time is a big performance hit.
+
+# xhprof_enabled: false  # Set to true to enable xhprof and "ddev start" or "ddev restart"
+# Note that for most people the commands
+# "ddev xhprof" to enable xhprof and "ddev xhprof off" to disable it work better,
+# as leaving xhprof enabled all the time is a big performance hit.
+
+# webserver_type: nginx-fpm, apache-fpm, or nginx-gunicorn
+
+# timezone: Europe/Berlin
+# This is the timezone used in the containers and by PHP;
+# it can be set to any valid timezone,
+# see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+# For example Europe/Dublin or MST7MDT
+
+# composer_root: <relative_path>
+# Relative path to the composer root directory from the project root. This is
+# the directory which contains the composer.json and where all Composer related
+# commands are executed.
+
+# composer_version: "2"
+# You can set it to "" or "2" (default) for Composer v2 or "1" for Composer v1
+# to use the latest major version available at the time your container is built.
+# It is also possible to use each other Composer version channel. This includes:
+#   - 2.2 (latest Composer LTS version)
+#   - stable
+#   - preview
+#   - snapshot
+# Alternatively, an explicit Composer version may be specified, for example "2.2.18".
+# To reinstall Composer after the image was built, run "ddev debug refresh".
+
+# nodejs_version: "18"
+# change from the default system Node.js version to another supported version, like 14, 16, 18, 20.
+# Note that you can use 'ddev nvm' or nvm inside the web container to provide nearly any
+# Node.js version, including v6, etc.
+
+# additional_hostnames:
+#  - somename
+#  - someothername
+# would provide http and https URLs for "somename.ddev.site"
+# and "someothername.ddev.site".
+
+# additional_fqdns:
+#  - example.com
+#  - sub1.example.com
+# would provide http and https URLs for "example.com" and "sub1.example.com"
+# Please take care with this because it can cause great confusion.
+
+# upload_dirs: "custom/upload/dir"
+#
+# upload_dirs:
+#   - custom/upload/dir
+#   - ../private
+#
+# would set the destination paths for ddev import-files to <docroot>/custom/upload/dir
+# When mutagen is enabled this path is bind-mounted so that all the files
+# in the upload_dirs don't have to be synced into mutagen.
+
+# disable_upload_dirs_warning: false
+# If true, turns off the normal warning that says
+# "You have Mutagen enabled and your 'php' project type doesn't have upload_dirs set"
+
+# working_dir:
+#   web: /var/www/html
+#   db: /home
+# would set the default working directory for the web and db services.
+# These values specify the destination directory for ddev ssh and the
+# directory in which commands passed into ddev exec are run.
+
+# omit_containers: [db, ddev-ssh-agent]
+# Currently only these containers are supported. Some containers can also be
+# omitted globally in the ~/.ddev/global_config.yaml. Note that if you omit
+# the "db" container, several standard features of ddev that access the
+# database container will be unusable. In the global configuration it is also
+# possible to omit ddev-router, but not here.
+
+# performance_mode: "global"
+# DDEV offers performance optimization strategies to improve the filesystem
+# performance depending on your host system. Should be configured globally.
+#
+# If set, will override the global config. Possible values are:
+#   - "global":  uses the value from the global config.
+#   - "none":    disables performance optimization for this project.
+#   - "mutagen": enables Mutagen for this project.
+#   - "nfs":     enables NFS for this project.
+#
+# See https://ddev.readthedocs.io/en/latest/users/install/performance/#nfs
+# See https://ddev.readthedocs.io/en/latest/users/install/performance/#mutagen
+
+# fail_on_hook_fail: False
+# Decide whether 'ddev start' should be interrupted by a failing hook
+
+# host_https_port: "59002"
+# The host port binding for https can be explicitly specified. It is
+# dynamic unless otherwise specified.
+# This is not used by most people, most people use the *router* instead
+# of the localhost port.
+
+# host_webserver_port: "59001"
+# The host port binding for the ddev-webserver can be explicitly specified. It is
+# dynamic unless otherwise specified.
+# This is not used by most people, most people use the *router* instead
+# of the localhost port.
+
+# host_db_port: "59002"
+# The host port binding for the ddev-dbserver can be explicitly specified. It is dynamic
+# unless explicitly specified.
+
+# mailhog_port: "8025"
+# mailhog_https_port: "8026"
+# The MailHog ports can be changed from the default 8025 and 8026
+
+# host_mailhog_port: "8025"
+# The mailhog port is not normally bound on the host at all, instead being routed
+# through ddev-router, but it can be bound directly to localhost if specified here.
+
+# webimage_extra_packages: [php7.4-tidy, php-bcmath]
+# Extra Debian packages that are needed in the webimage can be added here
+
+# dbimage_extra_packages: [telnet,netcat]
+# Extra Debian packages that are needed in the dbimage can be added here
+
+# use_dns_when_possible: true
+# If the host has internet access and the domain configured can
+# successfully be looked up, DNS will be used for hostname resolution
+# instead of editing /etc/hosts
+# Defaults to true
+
+# project_tld: ddev.site
+# The top-level domain used for project URLs
+# The default "ddev.site" allows DNS lookup via a wildcard
+# If you prefer you can change this to "ddev.local" to preserve
+# pre-v1.9 behavior.
+
+# ngrok_args: --basic-auth username:pass1234
+# Provide extra flags to the "ngrok http" command, see
+# https://ngrok.com/docs/ngrok-agent/config or run "ngrok http -h"
+
+# disable_settings_management: false
+# If true, ddev will not create CMS-specific settings files like
+# Drupal's settings.php/settings.ddev.php or TYPO3's AdditionalConfiguration.php
+# In this case the user must provide all such settings.
+
+# You can inject environment variables into the web container with:
+# web_environment:
+# - SOMEENV=somevalue
+# - SOMEOTHERENV=someothervalue
+
+# no_project_mount: false
+# (Experimental) If true, ddev will not mount the project into the web container;
+# the user is responsible for mounting it manually or via a script.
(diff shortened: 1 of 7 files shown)


apotek:
phpstan is throwing an error that I disagree with:

phpstan:
> Error: Variable $dbPath in isset() always exists and is not nullable.
> ------ ---------------------------------------------------------------- 
>  Line   src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php        
> ------ ---------------------------------------------------------------- 
>  140    Variable $dbPath in isset() always exists and is not nullable.  
> ------ ---------------------------------------------------------------- 

But `$dbPath` will certainly _not_ exist, if an exception occurs on line 133. So we need the `isset()` on line 140. 

```php
            try {
                $dbPath = $this->databaseDownload($siteName);
            } catch (TaskException $te) {
                $this->yell("$siteName: No database configured. Download/import skipped.");
                $result_data->append($te->getMessage());
                // @todo: Should we run a site-install by default?
                continue;
            }
            if (!isset($dbPath) || !is_string($dbPath) || strlen($dbPath) == 0) {
```

.lando.yml
adamzimmermann, line 1: 👏🏼

src/Robo/Task/Discovery/Alternatives.php
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Use this to figure out which binary to use from a series of alternatives.
+ *
+ * ``` php
+ * <?php
+ * // Figure out whether you should use more or less or cat on a system. The
+ * // result data will always contain a 'path' property, even if no alternative
+ * // could be resolved and the path is empty.
+ * $result = $this->task(Alternatives::class, 'more', 'less,cat')->run();
```
adamzimmermann:
```suggestion
 * $result = $this->task(Alternatives::class, 'more', ['less', 'cat'])->run();
```
I don't see the use of `explode` so I would expect an array, but perhaps I'm just not following this fully?

src/Robo/Task/Discovery/Tasks.php
```diff
@@ -0,0 +1,17 @@
...
+
+trait Tasks
+{
+    /**
+     * @param string $command
+     * @param array $alternatives
+     *
+     * @return \Usher\Robo\Task\Discovery\Alternatives|\Robo\Collection\CollectionBuilder
+     */
+    protected function taskAlternatives(string $command, array $alternatives)
```
adamzimmermann, line 13: I'm curious why you aren't using this trait in the `DevelopmentModeBaseCommands` class.

adamzimmermann (commented):
Some questions (mostly out of my own ignorance), but this is great work. 👏🏼 

Perhaps we can begin using this as is, but also begin working on trying to get something like this into Robo as a Plan B/long-term play?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -123,43 +123,61 @@ public function databaseRefreshLando(string $siteName = 'default'): Result
     /**
      * Refresh database on Tugboat.
      */
-    public function databaseRefreshTugboat(): Result
+    public function databaseRefreshTugboat(): ResultData
     {
         $this->io()->title('refresh tugboat databases.');
-        $result = null;
+        $result_data = new ResultData();
```
apotek: Make this camelCase

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -123,43 +123,61 @@ public function databaseRefreshLando(string $siteName = 'default'): Result
...
                 // @todo: Should we run a site-install by default?
                 continue;
             }
-            if (!is_string($dbPath) || strlen($dbPath) == 0) {
+            // phpstan doesn't seem to be able to understand that $dbPath will
+            // be undefined if an exception is thrown in try{}, so it complains.
+            // I could not find any identifier that would silence this so I had
+            // to silence the whole line :(.
+            /* @phpstan-ignore-next-line */
+            if (!isset($dbPath) || !is_string($dbPath) || strlen($dbPath) == 0) {
```
markdorison:
```suggestion
            if (empty($dbPath) || !is_string($dbPath) || strlen($dbPath) == 0) {
```
How about this?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -123,43 +123,61 @@ public function databaseRefreshLando(string $siteName = 'default'): Result
...
-        $result = null;
+        $result_data = new ResultData();
+
         foreach (array_keys($this->getAllSitesConfig()) as $siteName) {
             try {
                 $dbPath = $this->databaseDownload($siteName);
-            } catch (TaskException) {
+            } catch (TaskException $te) {
                 $this->yell("$siteName: No database configured. Download/import skipped.");
+                $result_data->append($te->getMessage());
```
markdorison:
```suggestion
                $resultData->append($te->getMessage());
```

[Symfony coding standards want camel case for variable names](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions).

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek:
```suggestion
            if (empty($dbPath) || !is_string($dbPath)) {
```

I don't think we need the strlen() anymore.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: Ok. I'll apply this throughout the function. Thank you.

src/Robo/Task/Discovery/Tasks.php
apotek: > I'm curious why you aren't using this trait in the `DevelopmentModeBaseCommands` class.

It's not a trait. It's a task.

.lando.yml
```diff
@@ -0,0 +1,9 @@
+name: usher
```
markdorison, line 1: @apotek Since we are including Lando config, do you mind if I push one up for DDEV?

apotek: > @apotek Since we are including Lando config, do you mind if I push one up for DDEV?

No problem at all! I guess I should have added this in a separate branch, but it was just for helping me develop so I left it here. But sure, a ddev would be helpful too.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: @markdorison with `empty($dbPath)` phpstan complains as follows:

> phpstan: Construct empty() is not allowed. Use more strict comparison.

And when I use `!isset($dbPath)` phpstan says:

> phpstan: Variable $dbPath in isset() always exists and is not nullable.

:(

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -123,43 +123,57 @@ public function databaseRefreshLando(string $siteName = 'default'): Result
...
                 $dbPath = $this->databaseDownload($siteName);
-            } catch (TaskException) {
+            } catch (TaskException $e) {
                 $this->yell("$siteName: No database configured. Download/import skipped.");
+                $resultData->append($e->getMessage());
                 // @todo: Should we run a site-install by default?
                 continue;
             }
-            if (!is_string($dbPath) || strlen($dbPath) == 0) {
+            if (!is_string($dbPath) || strlen($dbPath) === 0) {
```
apotek: Ok. So I can get around the fact that phpstan doesn't like `empty($dbPath)` and phpstan doesn't like `isset()` on a variable that (it thinks) is defined, though it is defined in a try and won't be defined in the case of a failure, by forcing the variable to be set to an empty string at top. PHPStorm doesn't like that either, but I can tell PHPStorm to be quiet. 

But now, Rector is complaining that my test for the empty string should not use `strlen()` but should use `empty()` and we know how phpstan feels about `empty()`

So phpstan hates `empty()` because it is vague, and Rector hates `strlen($var) === 0` and wants me to use `empty()` .... 

Making me think maybe I have to do a string to string comparison like `$dbPath === ''` which, I guess is ok, but feels wrong to me. 

so... spinning my wheels between these validators complaining.  Any suggestions?

^^ @markdorison

src/Robo/Task/Discovery/Tasks.php
adamzimmermann: 🙈

apotek: > 🙈

Super confusing. To add a task, you need to define a trait that is then used in the Robo file. The trait loads your task. It's a trip.

apotek:
> Some questions (mostly out of my own ignorance), but this is great work. 👏🏼
> 
> Perhaps we can begin using this as is, but also begin working on trying to get something like this into Robo as a Plan B/long-term play?

The "official" way to add tasks, is to simply put what I have in src/Robo/Task/Discovery into a php package with a robo-task type. So that would be an open source contrib and would then not need to be approved by Robo maintainers. But I suppose I could fork and make a merge request too. I'm open to it, if only because I think it would be interesting to see all the feedback about what I have done wrong (which would teach me more about Robo).

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: Ok. I'm just going to go with `$dbPath === ''`  I have this "sensation" as a developer that comparing a string to `''` is somehow more error prone than `strlen() === 0` or `empty()`. But I don't know why. 🤔 Maybe because to me a string with no characters is different than a quoted "nothing you can see here". 🩺

src/Robo/Task/Discovery/Tasks.php
adamzimmermann: Ok, that makes me feel better, as I was struggling to piece this all together.

adamzimmermann:
I'm going to defer final approval to @markdorison here. Looking good to me though.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -123,43 +123,57 @@ public function databaseRefreshLando(string $siteName = 'default'): Result
...
                 continue;
             }
             $dbName = $siteName === 'default' ? 'tugboat' : $siteName;
-            $result = $this->taskExec('mysql')
+            $taskResult = $this->task(Alternatives::class, 'mariadb', 'mysql')->run();
+            if (!$taskResult->wasSuccessful()) {
+                $resultData->append($taskResult);
+                continue;
+            }
+            $dbDriver = $taskResult->getData()['path'];
```
markdorison, line 152: There is only one instance of `$taskResult->getData()['path'];` in this PR, but I am imagining a future where this is used more commonly throughout the codebase. Would it be even better to have our method call to the `Alternatives` class return the path directly without having the caller need to know what to do with it and what array key to look for?

.ddev/config.yaml
```diff
@@ -0,0 +1,262 @@
+name: usher
+type: php
+docroot: ""
+php_version: "8.2"
+webserver_type: nginx-fpm
```
apotek, line 5: I went with Symfony for the lando config. Not sure if this matters at this stage of the game. But I can change the Lando one to be consistent with the ddev setup if you prefer.

apotek (commented):
@markdorison I left a question re Symfony versus php-fpm container style.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: This is a great question. I went back and forth on that. Your argument has the following merits:

1) The user of the Task needs only to call a self-documenting getter, like `getPath()`.
2) The user doesn't need to know or rely on the internals of the returned data. 

The drawbacks is:
1) `getData()` is a standard, documented Robo API for getting task data from a 
Task Result. So the user doesn't need to reach for a new method. They can just use getData() as they always do, and then dig into it to get their result. It's more convoluted, but transparent and doesn't require the user to learn a new API/method. 

In the end, I opted to simply go with the Robo Result interface and not futz with it. 

The best of both worlds: Write an AlternativesResult class that implements the Robo Result interface, and override the getData() method to just return the string. This way I am following the Robo API instead of inventing new stuff. But it's more boiler plate.

apotek: Looking into it more, the problem is that `getData()`, according to the interface, is supposed to return an array. Setting it to a string in a custom Result class would break the interface. 

So I could offer both: A convenience method for users who want one, and the regular route for people already accustomed to Robo and who will just fish for their data using `getData()` because that's what they are used to.

Either way, a Robo Task must return a Result object. But there's nothing stopping me from adding a `getPath()` method to the task. Inside that `getPath()` I could try to do some validation to ensure that `run()` has been called (either as an implied run() (ie, trigger a run() if run() has not been called), or as a check. I guess I could add the Result as a class property to the Task and in the `getPath()` method, see if that property has been set, and either just throw an exception if `run()` was not called first, OR, call `run()` and then return the path (probably with a parameter passed to it to force it). At this point though I think I have forced a user to have to learn too much about the task. 

Inside `Alternatives::run()` when we have a Result, before returning it set the `Alternatives::$result` class property to the result. 

Inside `Alternatives::getPath($force = FALSE)` check if `Alternatives::$result` is set and return just the path or throw an exception or, if $force is set to TRUE, execute `run()` and then return the path. 

Want me to do that @markdorison or just leave as is?

.ddev/config.yaml
markdorison: @apotek I went with `php` because [DDEV doesn't have a Symfony type](https://ddev.readthedocs.io/en/latest/users/configuration/config/#type). Not sure it matters.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
apotek: Or... is this the right time for me to get to know decorators better?

markdorison: @apotek You've laid it all out very well. What do you think we should do??

apotek: > @apotek You've laid it all out very well. What do you think we should do??

I messed around with decorators and, well of course even a decorator can't override an interface. A contract is a contract, so an array must be returned by getData(). 

That leaves option 2, which is putting a helper function in the Task itself rather than the Result object. But the pattern of Tasks is that they return a Result object and you work with that Result object, not that you run the task, ignore the Result (or sniff it) and then call some other method on the Task to get a specific kind of result. So, I _can_ implement a convenience method, but I feel that it goes against a pattern, is therefore unexpected, and would force a developer to study a README rather than do what they do with any other Robo task, which is look at the returned Result data via getData() and figure it out from there. 

So in the end, my answer is to leave it the way it is. ☘️

markdorison (approved):

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
markdorison: Sounds good!

PR #166, opened 2023-09-27
Fix invocation of Alternatives task

apotek:
Last minute refactor didn't get tested adequately. Second argument to alternatives task must be array.

Currently tugboat deployments with latest Usher will break. This merge request fixes the issue I introduced.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -11,7 +11,6 @@ use Rector\Set\ValueObject\SetList;
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->paths([
         __DIR__ . '/src',
-        __DIR__ . '/tests',
     ]);
 
     // Define sets of rules.
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -144,7 +144,7 @@ class DevelopmentModeBaseCommands extends Tasks
                 continue;
             }
             $dbName = $siteName === 'default' ? 'tugboat' : $siteName;
-            $taskResult = $this->task(Alternatives::class, 'mariadb', 'mysql')->run();
+            $taskResult = $this->task(Alternatives::class, 'mariadb', ['mysql'])->run();
             if (!$taskResult->wasSuccessful()) {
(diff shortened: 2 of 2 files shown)


apotek:
@adamzimmermann Once this is approved are you able to cut a new release that I can then update JFK code to use?

adamzimmermann:
@apotek the 8.1 test isn't passing. Can you take a look at that?

apotek:
> @apotek the 8.1 test isn't passing. Can you take a look at that?

Yes, I saw that last night but have no idea how to fix. The error is

> /home/runner/work/usher/usher/tests" does not exist.

But there is no usher/tests directory in the repo and I don't see any evidence of them having been there. 

The CI call is:

```
      - name: PHP rector
        run: composer rector
```

Composer rector resolves to:

```
"vendor/bin/rector process --dry-run",
```

Nothing in there is creating any directories or files. 

When I run it locally, it works

> % lando composer rector
> vendor/bin/rector process --dry-run
>  1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
> 
>  [OK] Rector is done!

adamzimmermann:
@markdorison I think we need to tag you in here to help us out.

apotek:
I can't reproduce the error locally. I'm going to guess this was a temporary github error and I am re-running the job.

apotek:
> I can't reproduce the error locally. I'm going to guess this was a temporary github error and I am re-running the job.

Nope. It failed again. 

So it looks like the command `rector process --dry-run` is unable to create some working directories it needs. 

However, when I run it in my container, I don't see any directories being created. I do see some tmp files being made. 

I have run all the CI commands locally in lando and they all work. 

This PR introduces no changes to composer.json or composer.lock so the tooling should be the same. Does github "cache" its worker containers in some way? Could it be that one was initialized poorly and is there a way to start from scratch?

markdorison (approved):

adamzimmermann (approved):

rector.php
apotek, line 1: I am very curious as to why this code didn't throw an error in my container but only in github's container.

PR #168, opened 2023-09-27
Allow projects to specify which user id to pass to drush for login.

markdorison:
## Description
Allow projects to specify which user ID to pass to drush for login.

## Motivation / Context
A project where uid 1 is blocked.

Tested locally.

## Screenshots
CleanShot 2023-09-27 at 13 03 40@2x

## Documentation
Updated `.sites.config.example.yml`

### DIFF
--- a/.sites.config.example.yml
+++ b/.sites.config.example.yml
@@ -2,6 +2,7 @@
 default:
   database_s3_bucket: S3_BUCKET_NAME
   database_s3_key_prefix_string: prod-db-prefix
+  drush_user_login_uid: 108
   theme_build:
     - theme_path: web/themes/custom/theme1
       theme_build_commands:
--- a/.sites.config.example.yml
+++ b/.sites.config.example.yml
@@ -2,7 +2,7 @@
 default:
   database_s3_bucket: S3_BUCKET_NAME
   database_s3_key_prefix_string: prod-db-prefix
-  drush_user_login_uid: 108
+  drupal_admin_uid: 108
   theme_build:
     - theme_path: web/themes/custom/theme1
       theme_build_commands:
--- a/.sites.config.example.yml
+++ b/.sites.config.example.yml
@@ -2,7 +2,7 @@
 default:
   database_s3_bucket: S3_BUCKET_NAME
   database_s3_key_prefix_string: prod-db-prefix
-  drupal_admin_uid: 108
+  drupal_user_login_uid: 108
   theme_build:
     - theme_path: web/themes/custom/theme1
       theme_build_commands:
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -192,7 +192,7 @@ class DevelopmentModeBaseCommands extends Tasks
         string $siteDir = 'default',
     ): Result {
         $this->io()->section("create login link.");
-        $uid = $this->getSiteConfigItem(key: 'drush_user_login_uid', siteName: $siteDir, required: false) ?? 1;
+        $uid = $this->getSiteConfigItem(key: 'drupal_admin_uid', siteName: $siteDir, required: false) ?? 1;
         return $this->taskExec($environmentType)
             ->arg('drush')
             ->arg("@$siteDir.$environmentType")
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -192,10 +192,12 @@ class DevelopmentModeBaseCommands extends Tasks
         string $siteDir = 'default',
     ): Result {
         $this->io()->section("create login link.");
+        $uid = $this->getSiteConfigItem(key: 'drush_user_login_uid', siteName: $siteDir, required: false) ?? 1;
         return $this->taskExec($environmentType)
             ->arg('drush')
             ->arg("@$siteDir.$environmentType")
             ->arg('user:login')
+            ->option("--uid=$uid")
             ->dir("$this->drupalRoot/sites/$siteDir")
             ->run();
     }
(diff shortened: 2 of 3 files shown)


.sites.config.example.yml
```diff
@@ -2,6 +2,7 @@
 default:
   database_s3_bucket: S3_BUCKET_NAME
   database_s3_key_prefix_string: prod-db-prefix
+  drush_user_login_uid: 108
```
apotek: Curious why this would be specific to drush. Wouldn't this be a general admin/utility account that one would use either with or without drush, for development and CI?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -192,10 +192,12 @@ public function drupalLoginLink(
         string $siteDir = 'default',
     ): Result {
         $this->io()->section("create login link.");
+        $uid = $this->getSiteConfigItem(key: 'drush_user_login_uid', siteName: $siteDir, required: false) ?? 1;
```
apotek: This `1` might be nicer as a class constant (`::DRUPAL_DEFAULT_ADMIN_UID`). Otherwise it appears as a magic number or truthy.

src/Robo/Plugin/Traits/SitesConfigTrait.php
```diff
@@ -78,12 +78,17 @@ protected function getSiteConfig(string $siteName = 'default'): array
...
-    public function getSiteConfigItem(string $key, string $siteName = 'default'): mixed
+    public function getSiteConfigItem(string $key, string $siteName = 'default', bool $required = true): mixed
     {
         $siteConfig = $this->getSiteConfig(siteName: $siteName);
         if (!isset($siteConfig[$key])) {
-            throw new TaskException($this, "Key $key not found for '$siteName' in $this->sitesConfigFile.");
+            if ($required) {
+                throw new TaskException($this, "Key $key not found for '$siteName' in $this->sitesConfigFile.");
+            }
+            return null;
```
apotek, line 91: This seems overly complicated. If we are offering contextual ability to ignore a missing config, then why not return an empty config when missing and just test for it/decide what to do, in the calling code? 

Alternatively, it seems if there really are properties that are "generally required" (ie, things will break without them), then that should be decided at the application level and not be able to be overridden by the caller. Hope this makes sense. Now that I read it, I'm not so sure 😆

apotek (commented):
Some thoughts unasked and offered anyway :).

src/Robo/Plugin/Traits/SitesConfigTrait.php
adamzimmermann: I'm not 100% sure I follow this, so consider me on the fence.

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
adamzimmermann: ++

adamzimmermann (approved):
Approving pending resolution of the discussions.

.sites.config.example.yml
markdorison: Would you just call it `user_login_uid`? I guess I imagined people wondering "user login for what?" Maybe I will go with `drupal_admin_uid`.

src/Robo/Plugin/Traits/SitesConfigTrait.php
markdorison: > This seems overly complicated. If we are offering contextual ability to ignore a missing config, then why not return an empty config when missing and just test for it/decide what to do, in the calling code?

I understand what you are driving at. It started this way, but then I had a bunch of code calling this method, and in most cases that code was all starting to look very similar with conditional checks and exceptions. My thought was I could simplify all of those calls and just do the check/exception here in `getSiteConfigItem`. I see the argument in both directions, but personally, I just got sick of writing the checks and exception code every time I wanted to call into site config. 

Another option would be to create two helper methods `getSiteConfigItemRequired` and `getSiteConfigItemOptional`, but that is just a different flavor of what we have here.

Thoughts?

> Alternatively, it seems if there really are properties that are "generally required" (ie, things will break without them), then that should be decided at the application level and not be able to be overridden by the caller.

Which application are we talking about here?

.sites.config.example.yml
```diff
@@ -2,6 +2,7 @@
 default:
   database_s3_bucket: S3_BUCKET_NAME
   database_s3_key_prefix_string: prod-db-prefix
+  drupal_admin_uid: 108
```
adamzimmermann: FWIW I liked the old variable name haha. This doesn't have to be an admin user. 🤷🏻‍♂️

src/Robo/Plugin/Traits/SitesConfigTrait.php
```diff
@@ -78,12 +78,17 @@ protected function getSiteConfig(string $siteName = 'default'): array
      *   The site configuration key to load.
      * @param string $siteName
      *   The site name.
+     * @param bool $required
+     *   Whether the config item is expected to always be present.
      */
-    public function getSiteConfigItem(string $key, string $siteName = 'default'): mixed
+    public function getSiteConfigItem(string $key, string $siteName = 'default', bool $required = true): mixed
```
adamzimmermann, line 84: Any concern with this breaking existing functionality? Maybe that's a feature not a bug though?

adamzimmermann, line 84: Any concern with this breaking existing functionality? Maybe that's a feature not a bug though?

adamzimmermann (approved):

src/Robo/Plugin/Traits/SitesConfigTrait.php
markdorison: I made the default `true` to mimic what existed previously.

.sites.config.example.yml
markdorison: Updated again!

src/Robo/Plugin/Traits/SitesConfigTrait.php
apotek: Following up here, @markdorison and I discussed defining a list of required configs and using that list in validation here. ie, if the config requested is empty and is in the required list, throw the exception, don't throw if not.

markdorison: I created https://github.com/ChromaticHQ/usher/issues/169 for this topic!

PR #170, opened 2023-10-06
Add additional value to twig.config debug settings.

markdorison:
## Description
Add additional value to twig.config debug settings.

## Motivation / Context
Match existing configuration on project site.

N/A

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
@@ -347,7 +347,7 @@ class DevelopmentModeBaseCommands extends Tasks
             );
         }
 
-        $this->say("enablig twig.debug in development services yml config.");
+        $this->say("optimizing twig for front-end development in development services yml config.");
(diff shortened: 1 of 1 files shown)


src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
```diff
@@ -347,11 +347,12 @@ protected function frontendDevEnableDrupal(string $siteDir = 'default', array $o
             );
         }
 
-        $this->say("enablig twig.debug in development.services.yml.");
+        $this->say("enablig twig.debug in development services yml config.");
```
apotek:
```suggestion
        $this->say("optimizing twig for front-end development in development services yml config.");
```

My thinking here is that we are changing three settings, not just debug. In your previous version, I _liked_ that the file was specified (though I suppose it should have been `fe.development.services.yml`) as opposed to making a general statement about the file. So... opinions.

apotek (approved):
I made a suggestion regarding output to the user, but the code will work as is.

PR #175, opened 2023-10-19
Remove Lando and Drupal 7 support.

markdorison:
## Description
- Remove Lando support.
- Remove Drupal 7 support.

I intend this to be included in a 4.0 release since we are dropping support for multiple tools. If/when this is ready to be merged, I will create a `4.x` branch and update the PR to point there. 

## Motivation / Context
Stay up-to-date with current needs.

Tested locally.

## Documentation
Updated.


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -77,12 +116,219 @@ public function devRefreshAll(
...
                 environmentType: LocalDevEnvironmentTypes::from($environmentType),
                 siteName: $siteName,
-                startLocalEnv: $startLocalDev,
             );
         }
         return $result;
     }
 
+    /**
+     * Refresh a site database in DDEV.
```
apotek, line 125: If we are moving to a model where there is only one supported local development model (ie DDEV), then I feel I would want to refactor this either as `databaseRefreshLocal` (removing implementation details from the name of the function as per line 134), or, better, by defining an interface with `databaseRefresh()` and then a base class and a class for each environment type (tugboat, local etc) that is loaded by a factory depending on environment.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -77,12 +116,219 @@ public function devRefreshAll(
...
+
+        $this->io()->section("importing $siteName database.");
+        $this->say("Importing $dbPath");
+        $this->taskExec(LocalDevEnvironmentTypes::DDEV->value)
+            ->arg('import-db')
+            ->option('database', $siteName == 'default' ? 'db' : $siteName)
+            ->option('file', $dbPath)
+            ->run();
+
+        // If a database was downloaded as part of this process, delete it.
```
apotek, line 153: Personally I get really frustrated if I waited a long time for a download and a script decided to delete it, meaning I have to re-download on my second invocation. Leave as-is, because I don't have a better idea. Other people might not like having multiple GBs of SQL dumps in their app. I wonder if we could come up with a pattern here (for later).

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -77,12 +116,219 @@ public function devRefreshAll(
...
+        if (!$dbPathProvidedByUser) {
+            $this->deleteDatabase($dbPath);
+        }
+
+        return $this->drushDeployWith(
+            localEnvironmentType: LocalDevEnvironmentTypes::DDEV,
```
apotek, line 159: Do we still need this? Seems like if we are defaulting to one local type, `drushDeployWith()` should always do a local deploy (ie currently ddev). Again, if we use an interface, base, and child pattern, this could be handled that way. Probably more of a refactor than desired for now.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -77,12 +116,219 @@ public function devRefreshAll(
...
+                $taskResult = $this->taskExec("$this->vendorDirectory/bin/drush")
+                    ->arg('cache:rebuild')
+                    ->dir("$this->drupalRoot/sites/$siteName")
+                    ->run();
+                $resultData->append($taskResult);
+            }
+        }
+
+        return $resultData;
+    }
```
apotek, line 218: I assume nothing really was changed here, just moved from the old file.

ibid with the below functions, all in green too.

apotek (approved):
This merge request appears to achieve the stated goal. 

It was difficult to parse out what has changed due to deletions and additions of files. But overall it looks sane and like it will work. 

I still think the whole approach here could be refined with a more pluggable architecture rather than multiple functions. It's probably not worth refactoring, since its an internal tool, but this is a great candidate for inheritance/sub-classing based on environment.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: I was inclined to leave the idea of supporting multiple environment types since it is already structured/written and who knows what the future holds. I was trying to avoid the pain of rewriting half of this 18 months from now. 

If we want to discuss excising even the _idea_ of supporting multiple local environments, maybe that can be its own ticket? How strongly do you feel about this?

src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php
adamzimmermann, line 1: So happy to see this file go away even though I created it!

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: When this was originally created, the databases used were small enough that redownloading was rarely an issue. Not deleting makes a lot of sense when we are dealing with very large databases. 

One concession I made towards this was only deleting the database if you are using one of the "wrapper" commands like this one, `database:refresh-ddev` or _its_ wrapper `dev:refresh`. If you call `database:download` directly the file will not be deleted, and then if you call these wrapper commands it will use that file assuming it is the latest database dump in S3.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: See above.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
markdorison: Correct.

adamzimmermann (approved):
🎉

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
apotek: This was a smart decision.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
apotek: > excising even the idea of supporting multiple local environments

I wasn't feeling strongly about it, but seeing the strong push to remove Lando everywhere made me think one way to dev was already the policy, so I was suggesting that the code no longer needed to have "flavors" in its function signatures.

markdorison:
Created 4.x branch and updated PR to point to that branch. Preparing to merge.

PR #205, opened 2024-03-20
Make refresh commands runnable inside container for ddev

### LINKED ISSUE #80: Update databaseRefreshDdev() to be able to be run from within the local dev container
## Description

[`databaseRefreshDdev()`](https://github.com/ChromaticHQ/usher/blob/e24704a2b41ab2b064e89d352e9f0294eada8ee6/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php#L62-L71) must be run from outside ddev since we [currently use the DDEV`db-import` command](https://github.com/ChromaticHQ/usher/blob/e24704a2b41ab2b064e89d352e9f0294eada8ee6/src/Robo/Plugin/Commands/DevelopmentModeBaseCommands.php#L81-L82) to handle the import. `robo dev:refresh` also uses this method so currently we are forced to run this from outside the container which means dependencies like PHP must be installed locally.

## Proposed Solution

Refactor `databaseRefreshDdev()` to be able to run from inside the ddev container.

## Additional Context

@goranmiric mentioned on a back-end call that he

### LINKED ISSUE #204: Fix DevelopmentModeCommands::databaseRefreshDdev() to be able to run inside docker container
## Description

The only thing blocking being able to run robo isolated from the vagaries of our various host OS's is the `DevelopmentModeCommands::databaseRefreshDdev` command. Why? Because it attempts to run the ddev command, which can only be run on the OS level. 

The fix here is to replace ddev import-db implementation with one that can run inside the container. 

For example, here's the drush way to do it, and it works just as well.

```
drush sql:drop --uri=SITENAME --yes
zcat "$file" | drush sql:cli --uri=SITENAME
```

And then do the same with the other `ddev drush...` incantations.

Making these changes will allow the command to run within the container, meaning that the entire spin up and refresh sequence can be run regardless of host environment php versions, composer versions, shell versions, other dependencies.

apotek:
As described in #204, we can't run robo refresh divorced from the host OS when using ddev because unfortunately the ddev-specific implementations are using ddev commands, which cannot be run inside the container. 

Removing the reliance on those commands and using native drupal/drush instead, allows us to run the robo refresh sequence entirely within the service container, meaning the developer doesn't have to futz with php versions, composer etc, on the host OS. 

Resolves #204

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -236,7 +236,7 @@ class DevelopmentModeCommands extends Tasks
     ): Result {
         $this->io()->section("create login link.");
         $uid = $this->getDrupalSiteAdminUid(siteName: $siteDir);
-        if ($environmentType == 'ddev') {
+        if ($environmentType === 'ddev') {
             return $this->taskExec('drush')
                 ->arg("@$siteDir.$environmentType")
                 ->arg('user:login')
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -325,7 +325,7 @@ class DevelopmentModeCommands extends Tasks
                 "'drush deploy' command not found. Further work is necessary to support this version of Drush."
             );
         }
-            // After drush deploy, re-import the latest configuration. This includes
+        // After drush deploy, re-import the latest configuration. This includes
         // the latest configuration_split configuration. Importing this twice
         // ensures that the latter command enables and disables modules based
         // upon the most up--to-date configuration. More at:
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -142,7 +142,7 @@ class DevelopmentModeCommands extends Tasks
             $dbPath = $this->databaseDownload($siteName);
         }
 
-        $this->io()->section("importing $siteName database.");
+        $this->io()->section("refreshing $siteName database.");
         $this->say("Dropping existing database for $siteName");
         $this->taskExec('drush')
             ->arg('sql:drop')
@@ -325,25 +325,18 @@ class DevelopmentModeCommands extends Tasks
                 "'drush deploy' command not found. Further work is necessary to support this version of Drush."
             );
         }
-        // After drush deploy, import the latest configuration again. This
-        // includes the latest configuration_split configuration. Importing this
-        // twice ensures that the latter command enables and disables modules
-        // based upon the most up--to-date configuration. Additional information
-        // and discussion can be found here:
+            // After drush deploy, re-import the latest configuration. This includes
+        // the latest configuration_split configuration. Importing this twice
+        // ensures that the latter command enables and disables modules based
+        // upon the most up--to-date configuration. More at:
         // https://github.com/drush-ops/drush/issues/2449#issuecomment-708655673
 
-        // ddev command is not available inside ddev containers.
-        if ($localEnvironmentType->value == 'ddev') {
-            return $this->taskExecStack()
-                ->dir("$this->drupalRoot/sites/$siteDir")
-                ->exec("drush @$siteDir.$localEnvironmentType->value deploy --yes")
-                ->exec("drush @$siteDir.$localEnvironmentType->value config:import --yes")
-                ->run();
-        }
+        // Currently we only have one local environment type:
+        // LocalDevEnvironmentTypes::DDEV, so no need for other implementations.
         return $this->taskExecStack()
             ->dir("$this->drupalRoot/sites/$siteDir")
-            ->exec("$localEnvironmentType->value drush @$siteDir.$localEnvironmentType->value deploy --yes")
-            ->exec("$localEnvironmentType->value drush @$siteDir.$localEnvironmentType->value config:import --yes")
+            ->exec("drush @$siteDir.$localEnvironmentType->value deploy --yes")
+            ->exec("drush @$siteDir.$localEnvironmentType->value config:import --yes")
             ->run();
     }
 
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -143,13 +143,14 @@ class DevelopmentModeCommands extends Tasks
         }
 
         $this->io()->section("importing $siteName database.");
-        $this->say("Importing $dbPath");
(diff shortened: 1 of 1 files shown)


apotek:
@mmatsoo / @mattdaniel Wondering if either of you have the bandwidth to look at this Usher change, allowing refresh to happen within the container instead of having to run it in the host OS context.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -235,6 +236,14 @@ public function drupalLoginLink(
     ): Result {
         $this->io()->section("create login link.");
         $uid = $this->getDrupalSiteAdminUid(siteName: $siteDir);
+        if ($environmentType === 'ddev') {
+            return $this->taskExec('drush')
+                ->arg("@$siteDir.$environmentType")
+                ->arg('user:login')
+                ->option("--uid=$uid")
+                ->dir("$this->drupalRoot/sites/$siteDir")
```
mmatsoo, line 235: @apotek Is this line needed? Running drush using the alias in `ln 241` should take care of hitting the correct site anyway? (ie. `drush @paypal.ddev user:login --uid=1`)

mmatsoo (approved):
👍 I left one comment for contemplation, but looks good.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
apotek: @mmatsoo Good catch and good question! I wasn't sure why we were specifying both `--dir` and a drush alias, since the drush alias should be sufficient. 

But those were both there in the original code, and I decided to focus on removing the `ddev` wrapper rather than change the implementation, since I was aware that this functionality is fundamental to a lot of developer workflows and I was afraid to make the change. It's already going to cause some grief when developers have to remember to run this in ddev instead of on their host OS. That will be a pain-point for a while, but will keep us moving in the right direction of no longer having to worry about the host OS php version or composer version or drush version at all.

apotek:
I am going to add this to devops agenda or tech back end meeting of some sort so I can broadcast the implications of this change.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -142,14 +142,15 @@ public function databaseRefreshDdev(string $siteName = 'default', array $options
...
+        $this->io()->section("refreshing $siteName database.");
+        $this->say("Dropping existing database for $siteName");
+        $this->taskExec('drush')
+            ->arg('sql:drop')
+            ->option('uri', $siteName)
+            ->option('yes')
             ->run();
-
+        $this->say("Importing $dbPath");
+        $this->_exec("zcat '$dbPath' | drush sql:cli --uri=$siteName");
```
apotek, line 152: Here's the big drawback with loading the db inside the container: We are effectively transferring the data over a virtual network interface from one container to another, which is probably noticeably slower than loading the data into the database from within the database container itself, which is what the import-db command does. 

Now, we could try to run this in the database container, but it would require mounting the benz-platform directory into the database container, which ddev does not do by default. 

To do this "right" we would have to:
1. Tell ddev to mount our application root directory into the database server somewhere, so that the database server has our db files accessible to it. 
2. Write a custom db command (not hard*) to load the data into the database. 

* There is _one_ part that is hard, and that is finding the correct authentication information. Granted, we can hard-code it in the command since it's always the same in ddev for all the databases, but the right way would be to get the connection info either by parsing the php settings, or (better) from `drush sql:connect` but if you run drush sql:connect, you need a fully working php/drupal environment installed into the database container. So that's a no-go. 

For these reasons, I went with just using drush, as above, even though it is transferring the data between containers.

apotek:
@mattdaniel I'd love your eyeballs on this at some point.

apotek:
Run a comparison on an in-container database load versus using ddev import-db for a site like JFK.

@mattdaniel Makes the point that implementing this does not remove the possibility of using the ddev native database import.

@agarzola It is important to engineer our tooling so that it can be more agnostic, more composable. So the less we build in assumptions about `lando` or `ddev` into our tooling and can create implementation-neutral tooling, the better. So our composer commands should prefer to not mention ddev.

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -315,16 +324,18 @@ protected function drushDeployWith(
...
-            ->exec("$localEnvironmentType->value drush @$siteDir.$localEnvironmentType->value deploy --yes")
-            // Import the latest configuration again. This includes the latest
-            // configuration_split configuration. Importing this twice ensures that
-            // the latter command enables and disables modules based upon the most up
-            // to date configuration. Additional information and discussion can be
-            // found here:
-            // https://github.com/drush-ops/drush/issues/2449#issuecomment-708655673
-            ->exec("$localEnvironmentType->value drush @$siteDir.$localEnvironmentType->value config:import --yes")
+            ->exec("drush @$siteDir.$localEnvironmentType->value deploy --yes")
+            ->exec("drush @$siteDir.$localEnvironmentType->value config:import --yes")
```
apotek, line 330: I would love help with insights into why we use `$localEnvironmentType->value` here. Shouldn't `drush @$siteDir` be sufficient? 

In other words, if `@$siteDir` and `$siteDir.ddev` are interchangeable, why use `$siteDir.ddev` at all? Simplicity would dictate to just use the site's name.

mmatsoo: My only guess is to support something like `@siteDir.lando` vs `@siteDir.ddev` vs `@siteDir.docksal` (I made that last one up...I think.)

apotek: Thank you @mmatsoo. Seems like we could someday probably get rid of all the environment type logic. I am not sure whether to do so in this merge request though, so am going to hold off and not refactor anything other than the exact thing we are trying to address.

apotek:
Going to hold off on merging this until the next devops council so we can broadcast the change and any potential fallout. Also not good to release this during time I am out of office.

apotek:
We need a rollout plan for this to minimize disruptions. 

1. When a site lead decided to update usher, what needs to be communicated to end users?
2. How to best add the .ddev/commands/web/robo command to each site?
3. Should this be a breaking usher version change? (ie, move from 4 to 5 and if we go from 4 to 5 should we also upgrade robo to v5?).

- [x] Investigate if robo5 introduces any breaking changes.
- [ ] Decide on including a ddev "robo" command in each repo or as a ddev add-on
- [ ] Come up with boiler plate "README" for the update that each tech lead can use to change their README andn communicate procedure changes into their own project's README (also with our clients). 

```sh
#! /bin/bash
## Description: Runs robo inside ddev
## Usage: robo
## Example: "ddev robo deploy:drupal"
./vendor/bin/robo $@
```

PR #208, opened 2024-05-27
Use sandbox mode for MariaDB imports

agarzola:
## Motivation / Context

Related to chromatichq/benz-tickets#3861.


agarzola:
Actually using sandbox mode failed, despite Tugboat reporting that it has the patched version of MariaDB. I’m closing this PR in favor of the workaround in #209.

PR #209, opened 2024-05-27
Circumvent MariaDB’s sandbox mode on database import

agarzola:
## Description

This PR is no longer necessary, but it’s staying open and in draft mode until after we update our project(s) to no longer reference this branch. Closing the PR would leave a very tempting “delete branch” button that would break our builds on any projects currently pointing to this branch.

## Motivation / Context

Related to chromatichq/benz-tickets#3861.

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -199,7 +199,11 @@ class DevelopmentModeCommands extends Tasks
                 ->run();
             $resultData->append($taskResult);
             $this->io()->section("import $siteName database.");
-            $taskResult = $this->taskExec("zcat $dbPath | $dbDriver -h mariadb -u tugboat -ptugboat $dbName")
+            # With `tail +2`, we are temporarily stripping out the first line
+            # of the database dump to remove a sandbox command that MariaDB
+            # does not understand.
+            # See: https://mariadb.org/mariadb-dump-file-compatibility-change/
+            $taskResult = $this->taskExec("zcat $dbPath | tail +2 | $dbDriver -h mariadb -u tugboat -ptugboat $dbName")
                 ->run();
             $resultData->append($taskResult);
             $taskResult = $this->taskExec('rm')->args($dbPath)->run();


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -199,7 +199,11 @@ public function databaseRefreshTugboat(): ResultData
                 ->run();
             $resultData->append($taskResult);
             $this->io()->section("import $siteName database.");
-            $taskResult = $this->taskExec("zcat $dbPath | $dbDriver -h mariadb -u tugboat -ptugboat $dbName")
+            # With `tail +2`, we are temporarily stripping out the first line
+            # of the database dump to remove a sandbox command that MariaDB
+            # does not understand.
+            # See: https://mariadb.org/mariadb-dump-file-compatibility-change/
+            $taskResult = $this->taskExec("zcat $dbPath | tail +2 | $dbDriver -h mariadb -u tugboat -ptugboat $dbName")
```
markdorison, line 206: Will this _always_ be there in any database dump or do we need to check for its presence?

apotek: There will not always be the new sandbox line in the dump file. Currently we only get those from Platform SH. So yes, if this should be universalized, there would have to be a check, something like:

(easier to do in shell rather than wrapping shell with php):
```
if zcat $dbPath | head -1 | grep --silent '^\/\*\!999999'; then
  zcat $dbPath | tail +2 | $dbDriver -h mariadb -u tugboat -ptugboat $dbName
else
  zcat $dbPath | $dbDriver -h mariadb -u tugboat -ptugboat $dbName
fi
```

agarzola: Thank you both for the astute feedback on this! After some discussion on the original ticket, we decided on upgrading `mariadb-client` in Tugboat after all, so we no longer need this PR. I’ll close it after our tooling is updated to no longer reference this branch.

agarzola:
This can now be closed!

PR #218, opened 2024-08-19
Force inclusion of symfony/yaml

apotek:
## Description

While usually usher is used together with Drupal, and most Drupal installations require symfony/yaml, best practice (and in order to pass php tests) would be for usher to include all critical dependencies explicitly. 

This merge request requires symfony/yaml to ensure its inclusion in the autoloader.


mmatsoo (approved):
👍

PR #221, opened 2024-09-16
Do not force a cache rebuild after a database refresh

### LINKED ISSUE #220: Tugboat deployments fail if a service is required from a module that is being enabled, because database refresh calls a cache rebuild
## Description

When a developer adds a dependency on a service that is supplied by an as-yet-not-enabled module, the way to handle the chicken-and-egg problem in a deployment is to write an update hook that force enables the module before anything else happens (since updb is the first deployment command and does not do a full bootstrap). 

However, our database:refresh-tugboat command (databaseRefreshTugboat()) calls a cache rebuild at the end of the database refresh sequence. A cache rebuild does do a full bootstrap so any code relying on a service from a module that has not been enabled, will end up causing the refresh command to fail. 

## Issue Owner
@apotek

### Steps
1. Add a dependency on a service provided by a module that is not (yet) enabled.
2. Deploy to Tugboat
3. See deployment fail with:

>  [Exec] Running zcat adobe-2024-09-16-03:05:23.sql.gz | /usr/bin/mariadb -h mariadb -u tugboat -ptugboat adobe
>  [Exec] Done in 23.383s
>  [Exec] Running rm 'adobe-2024-09-16-03:05:23.sql.gz'
>  [Exec] Done in 0.006s
>  [Exec] Running /var/lib/tugboat/vendor/bin/drush 'cache:rebuild' in /var/lib/tugboat/web/sites/adobe
> 
> In CheckExceptionOnInvalidReferenceBehaviorPass.php line 119:
>                                                                                
>   The service "config_events_subscriber" has a dependency on a non-existent s  
>   ervice "search_api_solr_admin.command_helper".                               
>                                                                                
>  [Exec]  Exit code 1  Time 0.668s
>  [notice] Stopping on fail. Exiting....
>  [error]  Exit Code: 1 
>  [error]    in task Robo\Task\Base\Exec 
>    
> 2024-09-16T15:45:44.525Z - Command Failed (Tugboat Error 1064): Exit code: 1; Command: $TUGBOAT_ROOT/vendor/bin/robo database:refresh-tugboat

### Environment
- Tugboat

apotek:
A cache rebuild forces a full bootstrap and any new code that relies on an unenabled module will fail the build.

Resolves #220

### DIFF
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -203,14 +203,6 @@ class DevelopmentModeCommands extends Tasks
             $resultData->append($taskResult);
             $taskResult = $this->taskExec('rm')->args($dbPath)->run();
             $resultData->append($taskResult);
-
-            if (!$this->drupalVersionIsD7($this->drupalRoot)) {
-                $taskResult = $this->taskExec("$this->vendorDirectory/bin/drush")
-                    ->arg('cache:rebuild')
-                    ->dir("$this->drupalRoot/sites/$siteName")
-                    ->run();
-                $resultData->append($taskResult);
-            }
         }
 
         return $resultData;


mmatsoo (approved):
👍

mmatsoo:
I only wonder if a cache rebuild should run in a separate step...so that it's still done, but not tied to the DB import.

mattdaniel (approved):
✅

PR #223, opened 2024-10-30
Allow aws commands to use aws environment variables for auth

### LINKED ISSUE #222: Database Download Trait does not respect AWS credentials in environment variables
## Description
The Database Download Trait uses the AsyncAWS library to do S3 database backup downloads. 

According to the library, AsyncAWS respects environment variables set, for authentication.

But Usher does not allow for this and forces use of credentials stored in `$HOME/.aws/credentials`

https://github.com/ChromaticHQ/usher/blob/eed0a3972dce93a8c2b824f775bf3a05847fecf5/src/Robo/Plugin/Traits/DatabaseDownloadTrait.php#L35-L42

[According to their documentation](https://async-aws.com/authentication/#how-authentication-works), the following sources, in order, should be consulted for credential configuration:

1. [Hard-Coded Configuration](https://async-aws.com/authentication/hard-coded.html)
2. [Environment Variables](https://async-aws.com/authentication/environment.html)
3. [WebIdentity](https://async-aws.com/authentication/web-identity.html)
4. [Credential and Configuration Files](https://async-aws.com/authentication/credentials-file.html)
5. [ECS Container Credentials](https://async-aws.com/authentication/ecs-container.html)
6. [EC2 Instance Metadata](https://async-aws.com/authentication/ec2-metadata.html)

In a perfect world, we should support all of these, in this order. Perhaps the perfect world is achievable and Usher just gets out of the way, but if not, we can at least, minimally, allow for 1, 2, and 4, in that order. 

This means that, before we run lines 35-42 of the above quoted code, we should:

1) Check to see if the client is already configured with authentication values.
2) Check to see if the required environment variables are set, and
3) Only if the above two conditions are false, continue to our current logic.

apotek:
## Description
Allow aws commands to use aws environment variables for auth
Do not assume credential file creation will have valid credentials. 
Begin to move away from deprecated ->io constructs.

Resolves #222

## Testing
Open your composer.json file. Find the "repositories" section and add the Usher repository:

```json
        {
            "type": "git",
            "url": "https://github.com/ChromaticHQ/usher.git"
        },
```
(nb: Make sure you add that comma at the end)

Then require this test branch:

```
ddev composer require "chromatic/usher:dev-222-database-download-trait-does-not-respect-aws-credentials-in-environment-variables"
```
Note: if you have already done this previously, and are coming back for more testing be sure to go to vendor/chromatic/usher in your local repo and do a `git pull` to get latest. 

Run commands you normally use in your client project.! Report failures here! These are just examples.

```
  database:download <site>     
  database:refresh-ddev <site>
  dev:refresh <site>
  frontend:dev-disable               <site>  
  frontend:dev-enable               <site>
  etc...
```
First, make sure there are no AWS environment variables set in your container. Also remove the .aws/credentials. You should have an authentication error and be prompted to enter your credentials. Enter bad credentials. You should be prompted to re-enter your credentials. This will keep going until a successful authorization or you cancel. 

Now, break your .aws/credentials by modifying key or token so that it won't work. You should see a failure again and a prompt to fix your credentials. 

Break your .aws/credentials again.  Now, however, create a .ddev/.env file and put GOOD  aws credentials in it. Restart your ddev project. Now you should see success and not be prompted to enter the aws credentials, even though your ./aws/credentials file contains a bad token or key. 

(The environment variables override the credentials file as per spec).

Now, break your env variables by breaking a key or token in your .ddev/.env file and restarting your ddev project. You should see an error about your AWS environment variables being bad and that you need to fix them. 

You can then decide to either remove your environment variables and use the ~/.aws/credentials, or (recommended) just use the .env file. 

Remember to remove your modifications to composer.json and composer.lock! Don't commit that change.

### Expected Results
1. If you have environment variables set up and they are wrong, robo should exit with a message about your environment variables.
2. If you have environment variables set up and they are correct, the download should work and there should be no prompts for anything.
3. If you don't have environment variables and no credentials file, you should have an authentication error and a prompt to configure your credentials.
4. If you don't have environment variables and you have a credentials file but the credentials are wrong, you should have an authentication error and be prompted to configure the credentials.
5. If you don't have environment variables and you have a credentials file and it has valid credentials, there should be no error, no prompt, and it should just download.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,7 +1,7 @@
 parameters:
-  checkMissingIterableValueType: false
   customRulesetUsed: true
   ignoreErrors:
+    - identifier: missingType.iterableValue
     - message: '#\Call to deprecated method io\(\) of class Robo\\Tasks#'
       paths:
         - 'src/'
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -80,7 +80,7 @@ class DevelopmentModeCommands extends Tasks
         ConsoleIO $io,
         string $siteName = 'default',
         array $options = ['db' => '', 'environment-type' => 'ddev'],
-    ): Result {
+    ): Result|ResultData {
         ['db' => $dbPath, 'environment-type' => $environmentType] = $options;
         return $this->devRefreshDrupal(
             io: $io,
@@ -108,7 +108,7 @@ class DevelopmentModeCommands extends Tasks
     public function devRefreshAll(
         ConsoleIO $io,
         array $options = ['skip-sites' => '', 'environment-type' => 'ddev']
-    ): Result {
+    ): Result|ResultData {
         ['skip-sites' => $skipSites, 'environment-type' => $environmentType] = $options;
         $siteNames = $this->getAllSiteNames();
         $result = null;
@@ -121,6 +121,10 @@ class DevelopmentModeCommands extends Tasks
                 environmentType: LocalDevEnvironmentTypes::from($environmentType),
                 siteName: $siteName,
             );
+            if ($result instanceof ResultData) {
+                $io->say("Cancelling the refresh for all sites.");
+                return $result;
+            }
         }
         return $result;
     }
@@ -297,7 +301,7 @@ class DevelopmentModeCommands extends Tasks
         LocalDevEnvironmentTypes $environmentType,
         string $siteName = 'default',
         string $databasePath = '',
-    ): Result {
+    ): Result|ResultData {
         $io->title('development environment refresh. 🦄✨');
         $this->taskComposerInstall()->run();
         // There isn't a great way to call a command in one class from another.
@@ -306,7 +310,10 @@ class DevelopmentModeCommands extends Tasks
         $this->taskExec("composer robo theme:build $siteName")
             ->run();
         $this->frontendDevEnable($io, $siteName, ['yes' => true]);
-        $this->databaseRefreshDdev($io, siteName: $siteName, options: ['db' => $databasePath]);
+        $result = $this->databaseRefreshDdev($io, siteName: $siteName, options: ['db' => $databasePath]);
+        if ($result instanceof ResultData) {
+            return $result;
+        }
 
         return $this->drupalLoginLink($io, $environmentType->value, $siteName);
     }
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -6,6 +6,7 @@ use DrupalFinder\DrupalFinderComposerRuntime;
 use Robo\Exception\TaskException;
 use Robo\Result;
 use Robo\ResultData;
+use Robo\Symfony\ConsoleIO;
 use Robo\Tasks;
 use Symfony\Component\Yaml\Yaml;
 use Usher\Robo\Plugin\Enums\LocalDevEnvironmentTypes;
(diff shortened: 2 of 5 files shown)


mmatsoo:
@apotek Your instructions all worked a charm until I got to the part to put good credentials into `.ddev/.env` and restart `ddev`. I restarted `ddev` and tried the robo command several times, even switching to a file called `env` instead of `.env` and I kept getting prompted to enter AWS credentials. (I tried this on my Imprivata instance.)

```
ddev robo database:download imprivata

database download.
==================

➜  'imprivata' database_s3_region not set. Defaulting to us-east-1.
➜  'imprivata' S3 bucket: imprivata-drupal-8
➜  'imprivata' S3 Key prefix: 'prod-imprivata'
➜  Connecting to S3...

                                                                                                                        
 [ERROR] HTTP 403 returned for "https://imprivata-drupal-8.s3.us-east-1.amazonaws.com/?list-type=2".                    
                                                                                                                        
         Code:    AccessDenied                                                                                          
         Message: Access Denied                                                                                         
         Type:                                                                                                          
         Detail:                                                                                                        
                                                                                                                        

➜  Unable to authenticate to AWS S3.
            You can either set your credentials as environment variables and try again,
            or you can continue by configuring your AWS credentials file.

 Do you wish to configure your AWS S3 credentials file? (yes/no) [yes]:
```
So, the part to enter credentials works great, but my `.env` file didn't seem to get recognized / used.

apotek:
> So, the part to enter credentials works great, but my `.env` file didn't seem to get recognized / used.

Does your .env file have the correct credentials, and is it formatted like this?
```
AWS_SECRET_ACCESS_KEY=#############################
AWS_ACCESS_KEY_ID=#####################
```

(ie, i have made the mistake of copying an ini file to the env file thinking).  

If all this checks out, I would be interested to see if your environment variables were actually set, ie:

```
ddev ssh
env | grep AWS
```

should output the correct credentials. Is that the case? And they should persist even after reboot.  Thank you for testing this. You'll have to do a git pull in vendor/chromatic/usher to get latest should you try again.

danarbello (approved):
Reviewed this synchronously

mmatsoo:
@apotek It turns out my `.env` file needed to be like the example you gave, large caps:

```
AWS_ACCESS_KEY_ID = XXXXXXXXXXXXXXXXXXXX
AWS_SECRET_ACCESS_KEY = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

What I had before, when it didn't work, looked like this:
```
[default]
aws_access_key_id = XXXXXXXXXXXXXXXXXXXX
aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
I initially removed `[default]`, but it still didn't work, so I made the keys large caps and then it began working.

danarbello:
@mmatsoo Interesting. It should be noted that Usher copies over the creds in the latter format:

```
[default]
aws_access_key_id = XXXXXXXXXXXXXXXXXXXX
aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

So, even Usher doesn't capitalize the variable keys 🤔

apotek:
> @mmatsoo Interesting. It should be noted that Usher copies over the creds in the latter format:
> 
> ```
> [default]
> aws_access_key_id = XXXXXXXXXXXXXXXXXXXX
> aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
> ```
> 
> So, even Usher doesn't capitalize the variable keys 🤔

The AWS credentials file is in a format called an "ini" file, where a configuration section is delineated by a `[section]` and the variables are defined under it. You'll be familiar with this format from php.ini settings when configuring a lamp stack for example. It's a pretty widespread format and there are even native php functions to read this format `parse_ini_file()`.

Environment variables are a completely different beast. There is, by the way no technical reason environment variables are all upper case; it's a convention, similar to naming constants in upper case too.  The only reason the environment vars have to be upper case when doing an environment config, is that the aws CLI as well as the aws-async php library that Usher uses, expects the environment variables to be upper case. 

The .env file and the credentials file are entirely different ways to set up your credentials for AWS and can't be interchanged.

PR #226, opened 2024-12-04
Use nodejs_version: auto in DDEV and standardize on .node-version file instead

### LINKED ISSUE #225: Use .node-version file to specify node version
## Description
Currently we specify the node version with 

```
nodejs_version: "18"
```

in the DDEV configuration. 

A more standard (and flexible approach) would be to add a .node-version file with the node version in it and change the DDEV config to use

```
nodejs_version: auto
```

## Issue Owner
@apotek

## Proposed Solution

Add .node-version file and change the nodjs_version value in DDEV to `auto`.

apotek:
This change makes sure we use the now-standard `.node-version` file to set the node version, since DDEV now reads that file if `nodejs_version` is set to `auto`.

Resolves #225


mmatsoo (approved):
👍 LGTM!

apotek:
Thank you @mmatsoo !

PR #235, opened 2025-03-25
Handle change in PHPUnit invocation syntax

### LINKED ISSUE #233: PHPUnit 10 no longer supports --verbose option
## Description
The `robo job:run-unit-tests` job calls PHPUnit with the following parameters:

```php
        return $this->taskExec('XDEBUG_MODE=coverage vendor/bin/phpunit --debug --verbose')->run();
```

This is not compatible with version 10 and up of PHPUnit. And these are the versions that come along with Drupal 11. 

https://github.com/sebastianbergmann/phpunit/blob/10.0.0/ChangeLog-10.0.md

## Issue Owner
@apotek 

## Proposed Solution
We should remove the `--verbose` flag.  

The authors seem to indicate that this is the "replacement" for --verbose:

`phpunit --log-events-verbose-text <OUTPUTFILE>`

We would have to add something that then outputs the contents of that file, as --verbose would have done. 

## Tasks

- [ ] TK

### LINKED ISSUE #234: Usher relies on phpunit to exist to run job:run-unit-tests, but does not install/require phpunit
## Description

Usher relies on phpunit to run some of its commands but does not install phpunit.

Technically, it should install its dependencies.

However, does it make more sense in this case for the including platform to require phpunit (and decide on versions, etc) and then for Usher to gracefully fail its `job:run-unit-tests` if phpunit does not exist or there is no configuration for it? 

## Issue Owner
@apotek 

### Steps
1. Git checkout usher.
2. Run composer install
3. Note that phpunit is not in the vendor directory. 
4. Run job:run-unit-tests and see command not found error.

## Expected Behavior

Either Usher should install phpunit (since it expects it) or it should gracefully fail the job:run-unit-tests command absent the necessary environment. 

## Tasks

- [ ] Decide what to do.
- [ ] Do it.

apotek:
## Description

This merge request allows for invoking `robo job:run-unit-tests` across PHPUnit version < 10 and > 10. PHPUnit 10 dropped support for the `--verbose` flag. 

## Motivation / Context
Resolves #233 
Resolves #234 

Require this branch of Usher in your project and run `robo job:run-unit-tests`.  It would be especially good to test this in a project that uses PHPUnit 9 (or lower) and one that uses PHPUnit v 10 (or higher).

### DIFF
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -40,6 +40,9 @@ class CICommands extends Tasks
      * Command to run unit tests.
      *
      * @aliases punit
+     *
+     * @todo Remove support for PHPUnit --verbose syntax in next major Usher
+     * version.
      */
     public function jobRunUnitTests(): Result
     {
--- a/src/Robo/Plugin/Commands/CICommands.php
+++ b/src/Robo/Plugin/Commands/CICommands.php
@@ -23,6 +23,11 @@ class CICommands extends Tasks
     protected const PHPCS_DEFAULT_PHP_VERSION = '8.1';
 
     /**
+     * The maximum PHPUnit version that allows the --verbose flag.
+     */
+    protected const PHPUNIT_MAX_VERSION_WITH_VERBOSE_FLAG = 9;
+
+    /**
      * RoboFile constructor.
      */
     public function __construct()
@@ -38,6 +43,16 @@ class CICommands extends Tasks
      */
     public function jobRunUnitTests(): Result
     {
+        exec("phpunit --version | awk '{print $2}'", $output, $result_code);
+        if ($result_code === 0) {
+            $major_version = (int) $output[0];
+            if ($major_version > self::PHPUNIT_MAX_VERSION_WITH_VERBOSE_FLAG) {
+                return $this->taskExec(
+                    'XDEBUG_MODE=coverage vendor/bin/phpunit --debug --log-events-verbose-text phpunit.log'
+                )->run();
+            }
+        }
+        // Default to old PHPUnit verbose flag syntax.
         return $this->taskExec('XDEBUG_MODE=coverage vendor/bin/phpunit --debug --verbose')->run();
     }
 
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,7 @@
     },
     "require-dev": {
         "phpstan/phpstan-strict-rules": "^1.5 || ^2.0",
-        "drupal/core": "*"
+        "drupal/core": "*",
+        "phpunit/phpunit": "*"
     }
 }


src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -38,6 +43,16 @@ public function __construct()
      */
     public function jobRunUnitTests(): Result
     {
+        exec("phpunit --version | awk '{print $2}'", $output, $result_code);
+        if ($result_code === 0) {
+            $major_version = (int) $output[0];
+            if ($major_version > self::PHPUNIT_MAX_VERSION_WITH_VERBOSE_FLAG) {
+                return $this->taskExec(
+                    'XDEBUG_MODE=coverage vendor/bin/phpunit --debug --log-events-verbose-text phpunit.log'
```
agarzola, line 54: Looks like this updated script writes logs to a file. Am I reading that correctly? If so, I have a few questions:

1. Is logging to a file fundamentally different from the way the `--verbose` flag works in < 10? (i.e. Does < 10 print logs to stdout?)
2. Assuming ≥ 10 with `--log-events-verbose-text` writes verbose logs to a file, is there a default log filename that we could resort to (in the interest of predictability), or does this flag _require_ that we pass a filename?
    - To be clear: if we are _required_ to pass a filename, I think `phpunit.log` is the best possible name we could use! The only thing that would trump its use, in my mind, would be if PHPUnit had a reasonably named default that was different.
3. If printing logs to a file instead of stdout is indeed different from < 10, it may be worthwhile mentioning this difference in behavior according to the version of PHPUnit in the README (and the release notes when the time comes, for that matter).

src/Robo/Plugin/Commands/CICommands.php
```diff
@@ -38,6 +43,16 @@ public function __construct()
...
+                )->run();
+            }
+        }
+        // Default to old PHPUnit verbose flag syntax.
         return $this->taskExec('XDEBUG_MODE=coverage vendor/bin/phpunit --debug --verbose')->run();
```
agarzola, line 59: I want to suggest simplifying this to _only_ support PHPUnit ≥ 10 and cutting a new major release of Usher to clearly indicate this breaking change. However, I can see how there may be practical reasons not to do this (e.g. it may be a while before all of our projects can get on Drupal 11 —and thus PHPUnit 10— and we want to be able to release new fixes and features within the 5.x branch for all of those projects to enjoy).

How do you feel about adding a `@todo` comment at the start of this function block to drop support for < 10 in the next major version? If the comment points to a ticket in this repository, even better!

agarzola (commented):
Code changes look great. I assume you will test this in a couple of our projects (each with a different version of PHPUnit) before cutting a new release, and that you can do so more efficiently and competently than I possibly could!

I am leaving some questions and suggestions, but the code changes here look great.

src/Robo/Plugin/Commands/CICommands.php
apotek: > Is logging to a file fundamentally different from the way the --verbose flag works in < 10? (i.e. Does < 10 print logs to stdout?)

The --verbose flag in versions < 10 outputs all info directly to stdout. The --log-events-verbose-text flag in versions 10+ writes to a file and _also_ stdout. I don't know why they engineered it this way (and I'm trying to not sink too much time into it since this is internal time on the margins). We are required to pass a file name. 

> If printing logs to a file instead of stdout is indeed different from < 10, it may be worthwhile mentioning this difference in behavior according to the version of PHPUnit in the README (and the release notes when the time comes, for that matter).

We don't mention phpunit at all in the README. In fact, the `job:run-unit-tests` command is not documented at all in the README. I would love to be more thorough but since I don't get a budget for working on our Devops tools any work I do here comes out of my general "Internal" time, which is. usually insufficient to do the work adequately. 

I will definitely mention this in the release notes, though :)

src/Robo/Plugin/Commands/CICommands.php
apotek: > How do you feel about adding a `@todo` comment at the start of this function block to drop support for < 10 in the next major version? If the comment points to a ticket in this repository, even better!

I will add a @todo. I opted to make this function version-agnostic because I have found that release management of our internal tooling takes more time than the code changes. Communicating cross team, encouraging adoption of a version, managing what gets fixed in old deprecated versions versus what is being fixed in the current branch is slow and cumbersome (and we don't even really have any third party users to speak of!). 

The approach I have pivoted to is to set the latest official release to one version behind the current release. This allows for testing via dependabot updates without marking a release official right away. And it means that I, instead of having to test each release on every platform that uses Usher by myself (again with no budget), can allow the dependabot process to be my test bed. I of course will do whatever rudimentary testing I can. In this case, I have a Drupal 11 branch in Benz that will allow me to test this release, and I can test the old version by using main Imprivata. Once I feel that I can build in tugboat on a branch with latest release, then I will usually, when I have time, create an official release.

apotek:
thank you for your comments and suggestions Alfonso. I have added the todo you requested and will mention the changes in the release notes. Just passing this back to you for anything else you might want to add and for approval. Thank you.

src/Robo/Plugin/Commands/CICommands.php
agarzola: We don’t need a whole section. Even something like this would be helpful to anyone wondering why Usher is producing a `phpunit.log` file seemingly out of nowhere:

> Note that logging works differently in different versions of PHPUnit. PHPUnit versions < 10 (required in Drupal < 11) log output to stdout only, while versions 10+ of PHPUnit (required in Drupal 11+) outputs to stdout _and_ a `phpunit.log` file. We recommend adding `phpunit.log` to your `.gitignore` when upgrading to Drupal 11.

agarzola (approved):
Approving this as-is.

PR #245, opened 2026-01-31
DNM, DISCUSS: Update 4.x from 5.x

apotek:
## Description
Updating 4.x with 5.x updates.

I am opening this as a way to discuss 1) IF we should be maintaining any older version of Usher at all, and 2) If so, what kinds of backports should go into a 4.x branch, and which shouldn't. (If they both end up being the same branch, we should just encourage people to use 5.x.... Something should be different (the node js and php compability perhaps?). 

I am very open to the idea that with our budget we probably should just not maintain any older versions.

## Motivation / Context
Maintain backwards compatibility for anyone still on 4.x

This would essentially make 4.x identical to 5.x so you could "test" by using `dev-5.x` in your require, but this merge request is just for discussing what it means to backport, what principles should apply for that, and if we should even do it.


agarzola (commented):
> I am very open to the idea that with our budget we probably should just not maintain any older versions.

I would only maintain the branches we actively use in the course of our client work. More importantly, if we do decide to maintain the `4.x` branch, I do not see the value in simply merging `5.x` into it. Presumably `5.x` was started due to breaking changes (that is assuming we are observing semver), so merging `5.x` into `4.x` would presumably introduce breaking changes into that branch, breaking the semantic versioning contract.

But maybe I am missing something?

apotek:
> I would only maintain the branches we actively use in the course of our client work.
100% agreed. Thank you for saying that.


> More importantly, if we do decide to maintain the `4.x` branch, I do not see the value in simply merging `5.x` into it. Presumably `5.x` was started due to breaking changes (that is assuming we are observing semver), so merging `5.x` into `4.x` would presumably introduce breaking changes into that branch, breaking the semantic versioning contract.
> 
> But maybe I am missing something?

Nope. You are right on the money (which is also why I prefixed this with DNM). This merge request was created to just force the conversation.

I'll survey the team to see if anyone is using anything < 4.x or < 5.x. And act accordingly.

PR #247, opened 2026-02-05
Potentially speed up db imports using drush sql:connect instead of drush sql:cli

### LINKED ISSUE #239: Use Drush’s sql:connect instead of sql:cli in DDEV database refresh
## Description

When running `ddev robo dev:refresh SITE`, the following warning is printed to the console:

> [warning] It is slow to pass large amounts of data via stdin to the sql:cli command. See the Examples at https://www.drush.org/latest/commands/sql_cli/ for an alternative using sql:connect.

This appears to relate to the `databaseRefreshDdev()` method, where we indeed [pass the database dump](https://github.com/ChromaticHQ/usher/blob/0d24d6cb26b82a8abf09df2c579a122628add5bb/src/Robo/Plugin/Commands/DevelopmentModeCommands.php#L178) via stdin to `drush sql:cli`:

```php
        $this->_exec("zcat '$dbPath' | drush sql:cli --uri=$siteName");
```

The warning message seems to suggest that something like this (untested, purely sample) code may be faster:

```php
        $this->_exec("zcat '$dbPath' > temp_dump.sql");
        $this->_exec("drush sql:connect --uri=$siteName < temp_dump.sql");
```

## Issue Owner

@agarzola

## Proposed Solution

Try using `sql:connect` instead of `sql:cli` to see if database imports run faster.

## Alternatives Considered

The status quo, which is effective though a bit slow.

apotek:
## Description
This changes the way data is imported during a dev db refresh from using STDIN pipe to load unzipped data into Drupal to reading an unzipped sql file into a db client.

This is supposed to be faster.

## Motivation / Context
Performance: Resolves #239

If we find that this greatly improves the import, we may expand this merge request to also include an optimization for the Tugboat refresh command.

### Performance
Before testing, do three refreshes using the 5.x or latest of usher and get a time to import average. In order to minimize external factors, close all running software other than DDEV, Terminal and OrbStack (or docker composer). Also, provide the db file, rather than downloading it each time from AWS to remove network latency from the testing.

Then require this branch and run three more and see if there is a statistically significant improvement. 

How to use this branch rather than the Usher release:

`ddev composer require "chromatic/usher:dev-239-sql-connect"`

There are further improvements in this branch that could also be tested:
1. Test while providing an already unzipped sql file on the command line
2. Test without providing a file (downloading from AWS)
3. Test while providing an invalid file, and see that we are now failing the task with an exception.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,8 @@
             "phpstan/extension-installer": true,
             "php-http/discovery": true,
             "drupal/core-composer-scaffold": true
-        }
+        },
+        "optimize-autoloader": true
     },
     "require-dev": {
         "phpstan/phpstan-strict-rules": "^1.5 || ^2.0",
--- a/composer.json
+++ b/composer.json
@@ -57,8 +57,7 @@
             "phpstan/extension-installer": true,
             "php-http/discovery": true,
             "drupal/core-composer-scaffold": true
-        },
-        "optimize-autoloader": true
+        }
     },
     "require-dev": {
         "phpstan/phpstan-strict-rules": "^1.5 || ^2.0",
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -257,23 +257,10 @@ class DevelopmentModeCommands extends Tasks
                 ->run();
             $resultData->append($taskResult);
             $io->section("import $siteName database.");
-            // Unzip the sql data file.
-            $taskResult = $this->taskExec('gunzip')
-                ->option('--force')
-                ->arg($dbPath)
+            $taskResult = $this->taskExec("zcat $dbPath | $dbDriver -h mariadb -u tugboat -ptugboat $dbName")
                 ->run();
             $resultData->append($taskResult);
-            // Strip off the .gz extension for import.
-            $importFile = substr($dbPath, 0, -3);
-            $taskResult = $this->taskExec($dbDriver)
-                ->option('-h', 'mariadb')
-                ->option('-u', 'tugboat')
-                ->option('-ptugboat')
-                ->arg($dbName)
-                ->rawArg("< $importFile")
-                ->run();
-            $resultData->append($taskResult);
-            $taskResult = $this->deleteDataFile($importFile);
+            $taskResult = $this->deleteDataFile($dbPath);
             $resultData->append($taskResult);
         }
 
--- a/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
+++ b/src/Robo/Plugin/Commands/DevelopmentModeCommands.php
@@ -138,6 +138,8 @@ class DevelopmentModeCommands extends Tasks
      *   The Drupal site name.
      * @option db
      *   Provide a path to a database dump to be used instead of downloading the latest dump.
+     *
+     * The data file to be imported must be either a gzipped or non-gzipped sql file.
      */
     public function databaseRefreshDdev(
         ConsoleIO $io,
@@ -167,6 +169,21 @@ class DevelopmentModeCommands extends Tasks
             }
         }
 
+        if (str_ends_with($dbPath, 'sql.gz')) {
+            $importFile = substr($dbPath, 0, -3);
+            $gzip = TRUE;
+        }
+        elseif (str_ends_with($dbPath, '.sql')) {
+            $importFile = $dbPath;
+            $gzip = FALSE;
+        }
+        else {
+            throw new TaskException(
+                $this,
+                'Data import file must either be a .sql file or a .sql.gz file.',
+            );
+        }
+
         $io->section("refreshing $siteName database.");
         $io->say("Dropping existing database for $siteName");
         $this->taskExec('drush')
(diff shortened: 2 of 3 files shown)


src/Robo/Plugin/Commands/DevelopmentModeCommands.php
```diff
@@ -167,18 +169,45 @@ public function databaseRefreshDdev(
...
+            }
+        }
+        else {
+            if ($gzip) {
+                $this->_exec("gunzip --force '$dbPath'");
+            }
+        }
+        $io->say('Importing data from: ' . $importFile);
+        $this->_exec('$(drush sql:connect --uri="' . $siteName . '") < "' . $importFile .'"');
+        if (!$dbPathProvidedByUser || ($dbPathProvidedByUser && $gzip)) {
```
agarzola: The `||` here already implies that `$dbPathProvidedByUser` is truthy, so we shouldn’t need to check it again. You know this, so I wonder if you are checking it for the sake of clarity, in which case I’d suggest adding a comment:
```suggestion
        // If the user did not provide the import file or they provided a
        // gzipped file, delete the import file.
        if (!$dbPathProvidedByUser || $gzip) {
```

agarzola:
@apotek Thanks for working on this!

From the testing instructions:

> Then checkout this branch […]

How do I do that from a project that uses Usher? I’m afraid I’m not as well-versed with Composer as I am with Yarn/npm. 😅

apotek:
> How do I do that from a project that uses Usher? I’m afraid I’m not as well-versed with Composer as I am with Yarn/npm. 😅

Good question!

`ddev composer require "chromatic/usher:dev-239-sql-connect"` should do the trick. I'll update instructions. 

I just made some further modifications and pushed them up.

apotek:
My test results on Imprivata. 

command used: `ddev robo database:refresh-ddev --db prod-imprivata-2026-02-04-20:39:19.sql.gz`

| Type | Just import Avg. | Total Avg. |
|--------|--------|--------|
| `sql:cli` | 00:03:49 | 00:04:21 |
| `sql:connect` | 00:00:40 | 00:01:21 | 

For just the import part, that's a 6.5x speedup. Now, that is just Imprivata. This might vary depending on database size etc, but I think the improvement is massive enough that any variations will be outstripped by the overall performance gain. 

I'll wait for your results @agarzola , but I think this bodes well for further refactoring to also improve Tugboat refreshes... but that should maybe be a separate merge request.

apotek:
I went ahead and added a similar modification for Tugboat refreshes. Will be interesting to do a comparison there....

Oh. That's interesting. A recent zcat-based import in Tugboat for Imprivata clocked in at 49 seconds. 

With this modification, I see pretty much the same: 50 seconds. 

So it doesn't seem like there is much difference for Tugboat, `zcat | mariadb` and `mariadb < file.sql` seem to perform equivalently. 🤔

composer.json
```diff
@@ -57,7 +57,8 @@
             "phpstan/extension-installer": true,
             "php-http/discovery": true,
             "drupal/core-composer-scaffold": true
-        }
+        },
+        "optimize-autoloader": true
```
agarzola: What does this do?

agarzola (approved):
Thanks for the instructions on how to install this branch on my project, @apotek. I tested this today and the bottom line is: No contest! The improvement in import times is absurd. Nice work!

On my multi-site project, using one of the larger databases, I first downloaded the latest DB dump and then ran this ten times total (five before switching to this branch and five after switching):

```
ddev robo database:refresh-ddev TENANT --db PATH_TO_FILE
```

Here are the times reported in the _Done in `XYZ`s_ immediately after the import:

## Using v5.1

- 65s
- 89s
- 105s
- 77s
- 91s

## Using this branch

- 9.81s
- 9.78s
- 9.83s
- 9.96s
- 9.95s

I didn’t test this in Tugboat, but the results you reported in the comments seem fine to me. I say ship it!

src/Robo/Plugin/Commands/DevelopmentModeCommands.php
agarzola: This was an old comment. Please disregard, as it is not applicable to the latest version of this branch. Sorry!

apotek: I love that you caught that! I didn't catch that until phpstan advised me. I think I was struggling to figure out the logic of when to delete the unzipped file and when not to. The version you commented on was my third iteration...  trying to come up with a way of presenting the logic that felt readable and understandable. Writing the condition as you (and phpstan) suggested felt like it was expressing something very different from what I was "thinking" but it turns out that it was "true" that the import file should always be deleted if the file was a gzip. It just didn't feel true.

composer.json
apotek: This makes autoload, once its classmap has been built, no longer check the file system for classes, meaning that the class loading is much much faster.

However, this doesn't belong in this merge request and I'm going to remove it. 

What we in general should be doing, in production, is adding `--optimize-autoload` to our `composer install` invocations. 

It's a weird tradeoff. For a contributed library, one _should_ be able to put "optimize-autoloader" in the library's composer.json, but it can also get in the way of development (adding a new class can cause problems etc).

agarzola: Interesting. Thank you for explaining this!

PR #248, opened 2026-02-23
fix: use explicit nullable types for PHP 8.4 compatibility

markdorison:
## Description
Fix implicitly nullable parameters in GitHubStatusTrait deprecated in PHP 8.4. Also add PHP 8.4 and 8.5 to the CI matrix.

## Motivation / Context
Fix deprecations and include latest PHP versions in testing.

CI should pass.

### DIFF
--- a/src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
+++ b/src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
@@ -74,7 +74,7 @@ trait DatabaseDownloadTrait
                 $objects = $s3->listObjectsV2($requestConfig);
             } catch (\Exception $e) {
                 $io->error($e->getMessage());
-                throw new AbortTasksException('Unable to access AWS S3. Giving up.');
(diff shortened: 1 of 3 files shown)


src/Robo/Plugin/Traits/DatabaseDownloadTrait.php
```diff
@@ -74,7 +74,7 @@ public function databaseDownload(ConsoleIO $io, string $siteName = 'default'): s
                 $objects = $s3->listObjectsV2($requestConfig);
             } catch (\Exception $e) {
                 $io->error($e->getMessage());
-                throw new AbortTasksException('Unable to access AWS S3. Giving up.');
+                throw new AbortTasksException('Unable to access AWS S3. Giving up.', $e->getCode(), $e);
```
apotek, line 77: Nice touch! Thanks for adding that.

apotek (approved):
Correctly marks the properties as nullable. Thank you for doing this. 

Passes tests.

PR #257, opened 2026-09-23
Generate PHPUnit coverage only if xdebug is enabled

### LINKED ISSUE #256: Only run PHPUnit with coverage if Xdebug is enabled
## Description

The job:run-unit-tests command sets XDEBUG_MODE to 'coverage' regardless of whether Xdebug is enabled or not. This generates a warning which often is output as an exit code of 1 (error) so that PHPUnit appears to fail even if all tests pass. 

## Issue Owner
@apotek 

## To Reproduce

Run `robo job:run-unit-tests` with Xdebug disabled. You'll see a warning about xdebug not being enabled, and the job will exit with error code 1 (unless you have configured PHPUnit to not fail on warning (in the xml)). 

## Expected Behavior

Running the unit tests should only try to generate a coverage report if xdebug is enabled.

## Tasks

- [ ] Devise a command line test for whether Xdebug is enabled. 
- [ ] Adjust the Robo command based on the command line test.

apotek:
## Description
This change tells PHPUnit to only create a coverage report if Xdebug is enabled. 

It also adds a feature to allow PHPUnit test filter patterns. 

PHP Rector also insisted on some extraneous changes in order to pass tests. 

## Motivation / Context
Resolves #256

Enable Xdebug. Run `robo job:run-unit-tests`. Coverage report should be generated.
Disable Xdebug. Run `robo job:run-unit-tests`. No coverage report should be generated. 

In either case, the job should exit successfully (unless there are actual PHPUnit errors)

Extra Credit: You can now also run phpunit with a filter. So, if you just want to run tests in my_module, you can do `robo job:run-unit-tests my_module` and PHPUnit will limit its tests to that module. You can also pass class names, etc. See PHPUnit docs for the `--filter` parameter for more info.


mmatsoo (approved):
👍 Changes look sane to me.

